PR #8, opened 2020-06-07
log: fix a deadlock caused by conflicting lock order

BurtonQin:
This patch fixes a possible deadlock caused by conflicting lock order in src/log.rs:

|clear_logs|end_record|
|---|---|
|overlays.write()||
||appending.write()|
|appending.write()//deadlock!||
||overlays.write()//deadlock!|

The fix is to lift the `overlays.write()` above `appending.write()` in `end_record()`.


arkpar:
Thank you for detecting this. Even though `clear_logs` is never called in parallel with `end_record` this might change in the future. Proposed fix is suboptimal though. Calling `to_file` is expensive and the overlay lock should not be taken at this time.

I've fixed `clear_log` in 87abacb84b48dcdf71d70dbbba872c9bd35daced instead.

PR #15, opened 2020-09-17
Value iterator

cheme:
I was looking this afternoon if there was a way to get a value iterator over a column and started this branch.

This should be mainly for maintenance (and I wanted to use it for some small column to but other db could be use instead).

Things here should almost work except for the last value table where current format of value makes it difficult to distinguish multipart values from other value (in the other table we rely on the fact that we always got length followed by an included key to say that [0, 0] is a non initialized value).

So this implementation would require changes to multipart value (I don't want to iterate on index), my personal feeling is that it is probably not worth it, but I open this PR as a question in case this functionality could be interesting (can be close).

### DIFF
--- a/src/table.rs
+++ b/src/table.rs
@@ -40,11 +40,13 @@
 // Partial entry (continuation):
 // [MULTIPART: 2][NEXT: 8][VALUE]
 // MULTIPART - Split entry marker. 0xfffe.
+// TODO ? MULTIPART_SPLIT - Split entry marker. 0xfffd.
 // NEXT - 64-bit index of the entry that holds the next part.
 // VALUE: The rest of the entry is filled with payload bytes.
 //
 // Partial entry (last part):
 // [SIZE: 2][VALUE: SIZE]
+// TODO ? MULTIPART_END - Split entry marker. 0xfffc.
 // SIZE: 16-bit size of the remaining payload.
 // VALUE: SIZE payload bytes.
 //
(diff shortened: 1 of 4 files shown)


arkpar:
What's the use case for the value-only iterator? I'd imaging key iterator to be much more useful.

src/table.rs
```diff
@@ -237,6 +274,11 @@ impl ValueTable {
 				return Ok(false);
 			}
 
+			if key.is_none() && &buf[0..2] == INIT {
```
arkpar, line 277: 0-sized tables and 0-size values are legitimate. `INIT` might match a 0-sized value.

cheme: I did notice that (see comment), this needs to change the format for multipart to work.

cheme:
Some collection traversal, but it is only used in some poc/branch, nothing that really matters.
After more thought, I believe for this use case I can also use some indexing trie.

> I'd imaging key iterator to be much more useful.

Yeah, I would have store the key with the value and then it becomes a key value iterator (hacky).

From reading readme and code  I thought all stored keys are already hashed (except with some flag for key that are already some hash), so no real way to do key iterator?

src/table.rs
arkpar: This is not a problem with multipart, but with fixed-size tables as well. 0-sized entries are allowed in fixed size tables.

cheme: Oh, I remember: the size for those entries does include the size the other fields, so I was thinking it could not be 0.

PR #17, opened 2021-01-25
Update Dependencies, switch memmap to memmap2 

gnunicorn:
refs https://github.com/paritytech/substrate/issues/7956

[Memmap is unmaintained](https://rustsec.org/advisories/RUSTSEC-2020-0077.html), fortunately a drop-in fork exists (memmap2). This PR switches the codebase to that crate and updates all dependencies to their latest version.

_Note_: as memmap is exposed in an external type, this change is a breaking major. Therefore, according to SemVer, the next version of this crate released should be `0.2.0` (at least).

_Note 2_: one main update that happened since the fork is [removing the winapi-dependency in favor of a localised equivialent](https://github.com/danburkert/memmap-rs/commit/37d18bbf421c4833c0f662f56889f83f7e72ad25).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -11,7 +11,7 @@ description = "Key-value database for the blockchain"
 [dependencies]
 log = "0.4.8"
 parking_lot = "0.10"
-memmap = "0.7"
+memmap2 = "0.2"
 blake2-rfc = "0.2.18"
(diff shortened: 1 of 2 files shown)


arkpar:
Published as 0.2.0
It should be safe to use in substrate

PR #18, opened 2021-02-04
Fix panic when asking for a salted hash of key with len != 32

gnunicorn:
[When upgrading substrate to latest [parity-db 2.0.0`](https://github.com/paritytech/substrate/pull/8015), we [saw a panic in the benchmarking](https://gitlab.parity.io/parity/substrate/-/jobs/801118#L2252) of code we haven't touched. A short investigation revealed that the [recently added key salting feature](https://github.com/paritytech/parity-db/commit/64389057d3db2a72f92e43259db1a0ad65c66632) doesn't properly deal with keys of length other than `32`. This PR fixes that behavior for both circumstances of the key being smaller then (our case) but also larger than 32 character.

I wanted to add tests, but noticed there aren't any and that a general infrastructure for them is missing.. (e.g. easily create a testable `Column`)

@arkpar may I ask you to release this as `2.0.1` soon, so I can merge the dependencies PR?!?

### DIFF
--- a/src/column.rs
+++ b/src/column.rs
@@ -15,6 +15,7 @@
 // along with Parity.  If not, see <http://www.gnu.org/licenses/>.
 
 use std::collections::VecDeque;
+use std::cmp::min;
 use std::sync::atomic::{AtomicU64, Ordering};
 use parking_lot::RwLock;
 use crate::{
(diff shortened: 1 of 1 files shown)


arkpar:
This only uses lower 32 bytes of the key. If two long keys share the same 32 byte prefix there will be a collision. I've implemented a better fix in https://github.com/paritytech/parity-db/commit/8ba14c7f8d62fbbe0eb93cb2df65a28befea06f3 and published it as `0.2.1`

PR #20, opened 2021-04-21
Review notes

arkpar:
This Issue is created to discuss some of the code issues/questions found by @cheme over his review of parity-db code.
Not to be merged as is.

### DIFF
--- a/README.md
+++ b/README.md
@@ -134,6 +134,9 @@ TODO seems interesting to indicate the transaction buffer:
 
 So query goes through shared overlay and file, and if write is enable, query also goes through writer.
 
+Question: I did not imediatly get the purpose of the three files. I am still wondering why we do not use the mmap. Is issue that mmap does not work in case of system crash?
+Actually would not make sense as the `write_plan` `complete_plan` do not seems to force flush (so the mmap support crash), so the point is probably performance?
+
 # Potential issues
 * Memory mapped IO won't be able to support 32-bit systems once the index grows to 2GB.
 * Size amplification. Index grow up to about 50% capacity before rebalance is triggered. Which means about 50% of allocated space is actually used for occupied index entries. Additionally, each value table entry is only partially filled with actual data.
--- a/README.md
+++ b/README.md
@@ -25,10 +25,37 @@ Database should be restored to consistent state if IO is interrupted at any poin
 # Implementation
 
 ## Data structure
-Data is organized into columns. Each column serving a particular type of data, e.g. state or headers. Column consists of an index and a set of 16 value tables for varying value size.  
+Data is organized into columns. Each column serving a particular type of data, e.g. state or headers. Column consists of an index (up to 255) and a set of 16 value tables for varying value size.  
 
 ### Index
-Index is an is mmap-backed dynamically sized probing hash table. Each entry is a page of 64 8-byte entries, making 512 bytes.  Each 64-bit entry contains 32 bits of value address, 4 bits of value table index and 28 bit value `c` derived from  `k`. `c` is computed by skipping `n` high bits of `k` and taking the next 28 bits.  `k` is 256-bit key that is derived from the original key and is uniformly distributed. `n` is current index bit-size. First `n` bits of `k` map `k` to a page. Entries inside the page are unsorted. Empty entry is denoted with a zero value. Empty database starts with `n` = 16, which allows to put just 240 bits of `k` in the value table. 
+Index is an is mmap-backed dynamically sized probing hash table. Each entry is a page of 64 8-byte entries, making 512 bytes. Each 8-byte entry contains 32 bits of value address, 4 bits of value table index and 28 bit value `c` derived from  `k`. `c` is computed by skipping `n` high bits of `k` and taking the next 28 bits. `k` is 256-bit key that is derived from the original key and is uniformly distributed. `n` is current index bit-size. First `n` bits of `k` map `k` to a page. Entries inside the page are unsorted. Empty entry is denoted with a zero value. Empty database starts with `n` = 16, which allows to put just 240 bits of `k` in the value table. 
+
+
+#### Review Notes
+
+TODO not up to date: value range is not 32 bit but 6 (64 entry per page) + n as there is no more possible value from indexing (from comment).
+Leaving 64 - 4 - (6 + n) of hash (initially 38 bit).
+Value adress being n + 6 + 4 as we need to cover all entries of similar size case.
+
+Still could do with just 32 bit packed in page to find over the packed bit allowing better paralelle instruction support.
+
+512o 64 address of 64bit so 64 * 8 byte. AKA CHUNK_LEN.
+64 entries per chunck aka CHUNK_ENTRIES.
+-> but find is done only on less than half: (28 bit value), would make sense to me to pack those, maybe make then 32 bit and reduce value address to 28 bit + 4 bit table.
+
+2^16 -> 65k pages ~ 32Megs ix.
+
+TODO indicates `n` is called `index_bits` in code.
+Also currently `n` maximum is 64.
+Plus n increment by factor 2: 32m -> 64 -> 128: so 64 seems really enough.
+
+So logic is as much key content as possible to allow early key mismatch detection.
+Starting at 2^16 index * 64 = 4M value (2^22 index that needs to be available among all values categories: 26 bit of adressing).
+Remaining 38 bit of key.
+To fit in 64 bit, we can raise n to (64 - 10) / 2 = 27 (37 bit value address, 2^33 value
+
+TODO check Be and Le usage with u64.
+
 
 ### Value tables
 Value table is linear array of fixed-size entries that can grow as necessary. Each entry may contain one of the following:
--- a/README.md
+++ b/README.md
@@ -1,3 +1,19 @@
+# Review
+
+I did add md paragraph called review and comments starting with [Review].
+I also propose some simple changes:
+- minor syntactic changes.
+- additional comment when it would have save me code reading (I did not try to document all, just what would have help me moving through code while reading).
+
+Key fix to 32 byte:
+- index.rs: we only need 8 start byte so most size fits.
+
+Chunk (page) fix to 512 byte (64 entries):
+I am not sure if the choice here is related to the page size or the number of entries (with faster search would it be good to use bigger page size).
+
+
+- TODO a word on salt?? -> in col used for hash
+
 # A database for the blockchain.
 
 ## **WARNING: PartyDB is still in development and should not be used in production. Use at your own risk.**
@@ -37,7 +53,8 @@ TODO not up to date: value range is not 32 bit but 6 (64 entry per page) + n as 
 Leaving 64 - 4 - (6 + n) of hash (initially 38 bit).
 Value adress being n + 6 + 4 as we need to cover all entries of similar size case.
 
-Still could do with just 32 bit packed in page to find over the packed bit allowing better paralelle instruction support.
+To optimize the page search, I could think of : https://event.cwi.nl/damon2009/DaMoN09-KarySearch.pdf.
+Probably using 16 (8 seems to small like 256 over 64, 1/4 chance to conflict) bit of our k bits to search first. (then we need to access remaining of page but only at a given entry).
 
 512o 64 address of 64bit so 64 * 8 byte. AKA CHUNK_LEN.
 64 entries per chunck aka CHUNK_ENTRIES.
@@ -52,9 +69,11 @@ Plus n increment by factor 2: 32m -> 64 -> 128: so 64 seems really enough.
 So logic is as much key content as possible to allow early key mismatch detection.
 Starting at 2^16 index * 64 = 4M value (2^22 index that needs to be available among all values categories: 26 bit of adressing).
 Remaining 38 bit of key.
-To fit in 64 bit, we can raise n to (64 - 10) / 2 = 27 (37 bit value address, 2^33 value
+Considering the key content should at least key 16 bit, then max size for n becomes (64 - 10 - 16) = 38 bit (and 48 bit value address).
 
-TODO check Be and Le usage with u64.
+About Be and Le usage with u64.
+-> key material extracted from hash as BE (first 64 bits). Make senses to me.
+-> entry written as LE?? not sure, it puts value index first? BE everywhere seem less error prone.
 
 
 ### Value tables
@@ -77,12 +96,43 @@ Page size of 64 index entries trigger a reindex once load factor reaches about 5
 ### Reindex
 When a collision can't be resolved, a new table is created with twice the capacity. Insertion is immediately continued to the new table. A background process is started that moves entries from the old table to the new. All queries during that process check both tables.
 
+#### Review
+
+TODO duplicate tables, they got different names depending on accesses so it is ok, TODO does it lock write? prev comment indicate not but double querying: insert in new.
+TODO check if there may be concurrency issue with the background process??? (not if write is done on previous table: seems ~??).
+
 ## Transaction pipeline
 On `commit` all data is first moved to an in-memory overlay, making it available for queries. The commit is then added to the commit queue. This allows for `commit` function to return as early as possible.
 Commit queue is processed by a commit worker that collects data that would be modified in the tables and writes it to the available log file. All modified index and value table pages are placed in the in-memory overlay. The file is then handled to another background thread that flushes it to disk and adds it to the finalization queue.
 Finally, another thread handles the finalization queue. It reads the file and applies all changes to the tables, clearing the page overlay.
 
 On startup if the log files exists they are validated for corruption and enacted upon the tables.
+
+### Review
+
+'data is first moved to an in-memory overlay' -> is it log 'end_record' ? (begin_record create a writer that end_record flush into in memory.
+'The commit is then added to the commit queue' -> 'log.to_file' in 'end_record' : actually append before, but under a lock. -> meaning commit queue is file for 'append'. (the higher index in sort, make sense).
+-> at end the append become the source for next step
+'Commit queue is processed by a commit worker that collects data that would be modified in the tables and writes it to the available log file' -> this should be working on the two log file that are not use for append: source is flushing dest is reading: fn 'flush_one'.
+'The file is then handled to another background thread that flushes it to disk and adds it to the finalization queue' -> fn 'read_next' (use a LogReader).
+
+So three file with same format, ordered by execution: appending -> flushing -> reading
+appending receive committed change first: lock on commit of the single writer
+reading is writing to backend: lock by backend thread.
+flushing is ???: queue to avoid reading locking appending probably actually would make sense if just a tmp ptr to file.
+
+TODO put those on function from log.
+
+TODO unclear why clear value not at same pace -> that is probably when we read an insert index we insert a clear value which in read_end will remove value from in memory overlay: seems inefficient (except to allow atomicity), but what happen when you clear something that was change a second time before read_next call. Excpet if state machine is tightly processed, seems racy at first.
+
+
+TODO seems interesting to indicate the transaction buffer:
+- first the LogWriter overlay, local to the write only thread (not shared).
+- second the shared overlay, it is a buffer to allow queueing changes.
+- third is the file queuing change (redundant with shared overlay)
+- last is the db tables/files.
+
+So query goes through shared overlay and file, and if write is enable, query also goes through writer.
 
 # Potential issues
 * Memory mapped IO won't be able to support 32-bit systems once the index grows to 2GB.
--- a/README.md
+++ b/README.md
@@ -14,6 +14,9 @@ I am not sure if the choice here is related to the page size or the number of en
 
 - TODO a word on salt?? -> in col used for hash
 
+
+Question: not sure about the use case of 'validate'.
+
 # A database for the blockchain.
 
 ## **WARNING: PartyDB is still in development and should not be used in production. Use at your own risk.**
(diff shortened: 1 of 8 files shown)


README.md
```diff
@@ -1,3 +1,42 @@
...
+
+### Key type
+
+- index.rs: we only need 8 start byte so most size fits.
+
+### Index
+
+Chunk (page) size to 512 byte (64 entries):
+I am not sure if the choice here is related to the page size or the number of entries
+(with faster search would it be good to use bigger page size).
```
arkpar, line 30: It's a minimum size which result in a low enough collision chance. It might be beneficial to use page size matching typical SSD IO page size indeed, if the entries in the chunk are ordered. I'd welcome such an experiment with a benchmark.

README.md
```diff
@@ -1,3 +1,42 @@
...
+
+### Hash
+
+Not really sure why we use salt option.
```
arkpar, line 34: That's to mitigate an attack has been demonstrated in the audit. Substrate user may brute force finding such runtime storage keys, that result in the database keys sharing bits required to overflow the chunk quickly, which leads to index size blowing up. Salting makes such an attack against the whole network impossible.

README.md
```diff
@@ -1,3 +1,42 @@
...
+
+### Validate
+
+not sure about the use case of 'validate', seems only to restart from crash.
```
arkpar, line 38: Before replaying the log we want to make sure it is not corrupted.

README.md
```diff
@@ -22,13 +61,65 @@ Transaction are applied atomically. Queries can't retrieve partially committed d
...
+
+I firstly misunderstood 'Durability' (IO failing not breaking db) for some consitency (call to commit means data is written in db).
+So there is many irrelevant notes in code.
+Call to commit does not actualy mean data is written/flushed to disk.
+- commit queue in db.rs only write to in memory queue.
+- mmap not having flush guarantee, this is something I am not really confident in.
+- flush worker also defers log flushing to file.
+
+So the db requires graceful shutdown.
+If it does not happen, state can be incorrect.
```
arkpar, line 74: Yeah, the fact that indexes and value tables are not `msync`/`fsync`ed after enacting the log (if sync mode is enabled) is a big overlook on my part. I'll add a separate issue for that. 
So currently, even if sync mode is not enabled, there's still consistency guarantee at the process level. I.e. graceful shutdown of the database process is not really required. It may be interrupted at any moment, and as long as OS propagates all the pending writes, database will be recovered with the log. However if the OS/power is shut down unexpectedly with some of the IO performed out of order, this may lead to issues indeed.

README.md
```diff
@@ -22,13 +61,65 @@ Transaction are applied atomically. Queries can't retrieve partially committed d
...
+All should be fine as long as all persistence is in a single db, but we should be strongly underline
+that commit does not flush to disk.
+So if process is sigkilled, the state would be consistent (but if another persistence source is used, it could be
+in a different state than this db).
+
+If full system is kill, mmap could not have flush: is it problematic?
+
+-> using a single db (no rocksdb in pkdot) seems rather prioritary.
+(IIUC, current commit semantic behave like rocksdb with 'DBOptions::manual_wal_flush')
+(which can be a gain for rocksdb too (but would require some periodic similar flush call).
```
arkpar, line 87: Yeah, the fact that indexes and value tables are not msync/fsynced after enacting the log (if sync mode is enabled) is a big overlook on my part. I'll add a separate issue for that.
So currently, as you've noted, even if sync mode is not enabled, there's still consistency guarantee at the process level. I.e. graceful shutdown of the database process is not really required. It may be interrupted at any moment, and as long as OS propagates all the pending writes, database will be recovered with the log. However if the OS/power is shut down unexpectedly with some of the IO performed out of order, this may lead to issues indeed.

README.md
```diff
@@ -22,13 +61,65 @@ Transaction are applied atomically. Queries can't retrieve partially committed d
...
+2^16 -> 65k pages ~ 32Megs ix size per column initially.
+
+TODO indicates `n` is called `index_bits` in code.
+Also currently `n` maximum is 64 - 10. (but we would want some at least 16 key length so would say 64 - 26 = 38 bit).
+Even 38 bit seems really enough.
+
+So logic is as much key content as possible to allow early key mismatch detection.
+Starting at 2^16 index * 64 = 4M value (2^22 index that needs to be available among all values categories: 26 bit of adressing).
+Remaining 38 bit of key.
+Considering the key content should at least key 16 bit, then max size for n becomes (64 - 10 - 16) = 38 bit (and 48 bit value address).
```
arkpar, line 117: These are all valid points. As `n` increases, so does the side and of the address, and the size of  `c` decreases. Eventually leading to more probing of the value table. Although the index will need to be really huge before this becomes an issue.

README.md
```diff
@@ -22,13 +61,65 @@ Transaction are applied atomically. Queries can't retrieve partially committed d
...
+
+About Be and Le usage with u64.
+-> key material extracted from hash as BE (first 64 bits). Make senses to me.
+-> entry written as LE?? seems like all code defaults to LE.
```
arkpar, line 121: Since it makes no difference from the search perspective. LE is a bit faster on modern hardware I believe

src/column.rs
```diff
@@ -143,9 +145,20 @@ impl Column {
...
 	}
 
+	// [Review] would it make sense to allow different hash than blake2b.
+	// -> most likely, we only use the 64 first bit and don't really need crypto
+	// resistance, or am I missing something.
+	// Actually this would be true if we did store the full key (but we store the
+	// hash).
+	// Note storing full key should probably be done by writing it into value
+	// (then there is the issue of having weak hash and a collision: requires
+	// to check the stored value or have an api that returns multiple matches).
```
arkpar, line 155: There's an option to store raw keys per column and avoid double hashing. But in generals, having fixed size keys made for simple design. Hashing allows the user to disregard any potential collisions, although by itself it is not enough to protect against a malicious attacker

In general, I'd prefer no to add features, such as alternative hashing functions, or arbitrary key sizes, unless they are really required by substrate. The focus here is to  maximise state database efficiency above all else, while keeping the rest reasonably fast. State is by far our largest data set.

src/column.rs
```diff
@@ -143,9 +145,20 @@ impl Column {
...
 	pub fn hash(&self, key: &[u8]) -> Key {
 		let mut k = Key::default();
 		if self.uniform_keys {
+			// [Review] TODO use Cow to avoid instantiation?.
```
arkpar, line 159: `Key` is a fixed size buffer. Instantiation should be free, as the zero-initialization should be optimized away.

src/column.rs
```diff
@@ -278,18 +296,25 @@ impl Column {
 				}
 			};
 
-			if let Some((table, sub_index, existing_tier, existing_address)) = existing {
+			if let Some((table, sub_index, existing_tier, existing_address)) = existing_index {
 				let existing_tier = existing_tier as usize;
 				if self.collect_stats {
+					// [Review] seems rather costy for the unsized value. Could consider storing size for
+					// those multipart value.
```
arkpar, line 303: `collect_stats` is currently enabled by default, but the plan is to disabled it once the database is stable and all the values are tweaked. So this should not be an issue.

src/column.rs
```diff
@@ -278,18 +296,25 @@ impl Column {
...
 					let cur_size = tables.value[existing_tier].size(&key, existing_address.offset(), log)?.unwrap_or(0);
 					self.stats.replace_val(cur_size, val.len() as u32);
 				}
 				if self.ref_counted {
+					// [Review] no check value is equal: means ref_count only works when preimage is true?
+					// should we 'debug_assert!(target_tier == existing_tier);'?
```
arkpar, line 309: Instead I'd just document that inserting an existing ref-counted value will bump the counter and ignore the new value. Or maybe provide an additional `inc_ref` function to avoid using `insert`. Substrate wrapper around the database API already has similar semantics.

src/column.rs
```diff
@@ -314,6 +339,8 @@ impl Column {
 					PlanOutcome::NeedReindex => {
 						log::debug!(target: "parity-db", "{}: Index chunk full {}", tables.index.id, hex(key));
 						Self::trigger_reindex(tables, reindex, self.path.as_path());
+						// [Review] Here we will call write_insert_plan a second time which seems fine, but
+						// could be skipped by just calling again tables.index.write_insert_plan?
```
arkpar, line 343: I believe that it is theoretically possible that `write_insert_plan` might return `PlanOutcome::NeedsReindex` the second time. Reindexing begins in the background so the chunk may be full again by the second call.

src/column.rs
```diff
@@ -442,6 +478,8 @@ impl Column {
 	}
 
 	pub fn reindex(&self, log: &Log) -> Result<(Option<IndexTableId>, Vec<(Key, Address)>)> {
+		// [Review] is next TODO still relevant?
```
arkpar, line 481: Probably not

src/column.rs
```diff
@@ -395,10 +423,18 @@ impl Column {
...
+					// [Review] could this case be handle by triggering reindex in 'replay_all_logs',
+					// or is it specific to validate?
+
 					// Re-launch previously started reindex
 					// TODO: add explicit log records for reindexing events.
+
+					// [Review] I did think of writing it in log, but then conclude it would be wrong
+					// because reindex is triggered after enact_log. (so would need specific log).
+					// Would maintaining a persistence of 'index_bits' be enough (so we restart reindex at start)
+					// when opening index tables.
```
arkpar, line 436: Maintaining it where? We'd need to provide the same consistency guarantees as we do for log records.

src/column.rs
```diff
@@ -395,10 +423,18 @@ impl Column {
 					tables.index.validate_plan(record.index, log)?;
 				} else if let Some(table) = reindex.queue.iter().find(|r|r.id == record.table) {
 					table.validate_plan(record.index, log)?;
-				}
-				else {
+				} else {
+					// [Review] could this case be handle by triggering reindex in 'replay_all_logs',
+					// or is it specific to validate?
```
arkpar, line 428: We need to to start re-indexing during validation, so that validation can continue.

src/db.rs
```diff
@@ -30,6 +30,20 @@
 /// Each background worker is signalled with a conditional variable once
 /// there is some work to be done.
 
+// [Review] I am not sure of the use of the `flush_worker`: 
+// it is waiting of `Read` to be empty, so wouldn't it be more
+// straight forward to flush (`fsync`) in the commit_worker.
```
arkpar, line 35: The idea is that flushing logs can be done in the background and it should not block anything else really

src/db.rs
```diff
@@ -437,6 +463,9 @@ impl DbInner {
 						match reader.next()? {
 							LogAction::BeginRecord(_) => {
 								log::debug!(target: "parity-db", "Unexpected log header");
+								// [Review] on hall those validation error handling, 
+								// should we just crash the application. I mean
+								// is it ok to clear log silently?
```
arkpar, line 468: There's an issue here indeed. If the log is detected to be invalid, we should really enact all the valid logs up to this point to recover to a consistent state.

src/db.rs
```diff
@@ -519,6 +549,10 @@ impl DbInner {
 		if let Some((record_id, cleared, bytes)) = cleared {
 			self.log.end_read(cleared, record_id);
 			{
+				// [Review] when validation_mode is true we still process
+				// the queue. Logic here is that validation mode is not use
+				// from queue but on restart, maybe consider renaming 'validation_mode'
+				// to 'replay_mode'.
```
arkpar, line 555: > when validation_mode is true we still process the queue

Not sure what you mean here. `validation_mode` The commits that are processed with `validation_mode` are not in the commit queue.

src/db.rs
```diff
@@ -543,7 +577,16 @@ impl DbInner {
 	}
 
 	fn replay_all_logs(&self) -> Result<()> {
+		// [Review] should we also replay reindex first?
+		// ReIndex would be here in column from reading the file system,
+		// but the reindex worker may be stuck on the condvar?
+		// call to start_reindex should be costless here (even if record_id
+		// is not easy to find but putting it to 1 should be enough).
```
arkpar, line 584: Right, this looks like an issue.

src/index.rs
```diff
@@ -26,7 +26,7 @@ use crate::{
 
 const CHUNK_LEN: usize = 512; // bytes
 const CHUNK_ENTRIES: usize = 64;
-const HEADER_SIZE: usize = 512;
+const HEADER_SIZE: usize = 512; // [Review] Nothing in header??
```
arkpar, line 29: Reserved for future use ;-)

src/db.rs
cheme: I was mentioning the items from the log (referring to the log as queue was incorrect). Just the fact that the expresion folowing 'validation' if branch are processed (no 'else' statement).

src/index.rs
```diff
@@ -82,6 +83,8 @@ impl Entry {
 	}
 }
 
+// [Review]: #[repr(transparent)] of any use?
+/// Value address.
```
arkpar, line 87: As I understand, `repr(transparent)` concerns C ABI compatibility, which is not an issue here.

src/column.rs
cheme: Oh, my understanding on this point was that the function that trigger reindexing did switch the indexing to a new file with the new index bits.
Then write index is replayed and target the new index file.
While in background the reindex process just consume old data and write it to new file (which could trigger a reindex because of concurrent writes).
So even if all reindexing run before this log enactment, the size will only be half.

src/index.rs
```diff
@@ -203,6 +215,10 @@ impl IndexTable {
 		}))
 	}
 
+	// [Review] lazy creation, I don't think it is worth mentioning, pretty obvious
+	// by reading the type of map.
+	// Note that having a directory abstraction could ensure two identical path
+	// at not use at small cost.
```
arkpar, line 221: Not sure what you mean here. `id` should result in a unique filename, so there should be no path collisions.

src/index.rs
```diff
@@ -274,6 +299,9 @@ impl IndexTable {
 		if let Some(map) = &*self.map.read() {
 			let source = Self::chunk_at(chunk_index, map);
 			chunk.copy_from_slice(source);
+			// [Review] do not work if using BE target. Probably worth it to handle the case, I remember
+			// a PR for scale doing it (not super usefull but would avoid bounty discussion on it).
```
arkpar, line 303: Good point

src/index.rs
```diff
@@ -284,13 +312,17 @@ impl IndexTable {
...
+		// in other sub_index calls, sub indexs is a starting query point, here it is insertion point.
 		sub_index: Option<usize>,
 		log: &mut LogWriter,
 	) -> Result<PlanOutcome> {
 		let chunk_index = key >> (64 - self.id.index_bits());
 		if address.as_u64() >= (1 << Entry::address_bits(self.id.index_bits())) {
 			// Address overflow
 			log::warn!(target: "parity-db", "{}: Address space overflow at {}: {}", self.id, chunk_index, address);
+			// [Review] Shouldn't it be unreachable (would mean all chunk full and one triggering reindex
+			// before)?
```
arkpar, line 325: I think I've experimented at some point with having `address_bits` = `index_bits`+ 6. So that if value size is  distributed unevenly, we may end up overflowing a single value table and run out of address space before running out of index space. This should not be possible with the current setup indeed, but I'd leave it there in case we decide to tweak these values again.

src/index.rs
```diff
@@ -299,9 +331,12 @@ impl IndexTable {
 		let new_entry = Entry::new(address, partial_key, self.id.index_bits());
 		if let Some(i) = sub_index {
 			let entry = Entry::from_u64(u64::from_le_bytes(chunk[i * 8 .. i * 8 + 8].try_into().unwrap()));
+			// [Review] Can it be just debug_assert?
 			assert!(entry.key_material(self.id.index_bits()) == new_entry.key_material(self.id.index_bits()));
 			&mut chunk[i * 8 .. i * 8 + 8].copy_from_slice(&new_entry.as_u64().to_le_bytes());
 			log::trace!(target: "parity-db", "{}: Replaced at {}.{}: {}", self.id, chunk_index, i, new_entry.address(self.id.index_bits()));
+			// [Review] is changing the whole chunk necessary: we just need individual change over in
+			// memory log.
```
arkpar, line 339: It's just easier to implement a page overlay. If we move to k-ary search or some other form of sorted chunks, it will probably be necessary to overwrite the whole page anyway.

src/index.rs
```diff
@@ -387,6 +430,9 @@ impl IndexTable {
 		let offset = META_SIZE + index as usize * CHUNK_LEN;
 		// Nasty mutable pointer cast. We do ensure that all chunks that are being written are accessed
 		// through the overlay in other threads.
+
+		// [Review] most api uses '&self', seems like we could use &mut some place and remove
+		// some atomics, also not sure why we don't get write lock here.
```
arkpar, line 435: We don't want to lock all of the mapped memory and block reads for performance reasons. There needs to be synchronisation on the page level.

src/log.rs
```diff
@@ -27,6 +27,12 @@ use crate::{
 };
 
 pub struct InsertIndexAction {
+	// [Review] LogAction  and insert action are written
+	// from commit transacion (a record_id), all in memory.
+	// We could also order the content by column
+	// and use a start_column and end_column tag into a
+	// record instead: would be more compact.
+	// Would need that ChangeSet would need a btreeset.
```
arkpar, line 35: Might be worth it. Requires benchmarking though

src/log.rs
```diff
@@ -477,8 +508,13 @@ impl Log {
 		Ok(bytes)
 	}
 
+	// Read from queue log enacted to db successfully,
+	// we therefore can remove them from the shared overlay.
 	pub fn end_read(&self, cleared: Cleared, record_id: u64) {
+		// [Review] this can happen because this is record_id from
+		// db queue and db_queue can fail so some id can be skippend.
 		if record_id >= self.next_record_id.load(Ordering::Relaxed) {
+			// [Review] racy?
```
arkpar, line 517: I think this may only happen when replaying the log initially. This is used to initialise `next_record_id` from an existing log.  After that it increments normally in `begin_record`. Not the most elegant solution and perhaps could be better solved by pre-scanning the log.

src/log.rs
```diff
@@ -534,6 +574,14 @@ impl Log {
...
 			}
 
+			// [Review] this clear reading file: but we
+			// do not know that mmap(s) got flushed before doing it.
+			// (I would keep it until flush, but do not know how to
+			// listen to mmap flush).
+			// Guess we could ensure buy usingf MAP_SYNC MAP_SHARED_VALIDATE.
+			// -> mmap2 is oppening MAP_SHARED so no MAP_SYNC at this point.
+			// But still MAP_SYNC looks a bit frightening, quote from https://lwn.net/Articles/731706/:
+			// `The result is a relatively simple mechanism that will perform far better than the currently available alternative — manually calling fsync() before each write to persistent memory. The potential downside is that any write operation can now create a flurry of I/O as the filesystem flushes out dirty metadata. That can cause the process to block in what was supposed to be a simple memory write, introducing latency that may be unexpected and unwanted.`
```
arkpar, line 584: The important bit is not to delete the `reading` file until flushing is complete. I suppose we could simply synchronously call `Mmap::flush` and `File::sync_data` for all indexes/tables here. It should not block other IO. 
`MAP_SYNC` does not look right indeed.. We don't want each write to be synced immediately, but rather sync at transaction, or even log file boundary.

src/table.rs
```diff
@@ -54,6 +58,8 @@
 // TOMBSTONE - Deleted entry marker. 0xffff
 // NEXT - 64-bit index of the next deleted entry.
 
+// [Review] for big value, would it be an idea to directly store a file instead of miltiparting.
+// Like using NEXT pointing to 0 (metadata).
```
arkpar, line 62: I'm afraid this can lead to lots of files in some extreme cases.Typically we don't need that much performance when reading/writing big blobs (think. block bodies)

src/table.rs
```diff
@@ -247,6 +258,9 @@ impl ValueTable {
 			};
 
 			if part == 0 {
+				// [Review] Skipping 6 first byte does not make much sense to me.
+				// It is used to build key material, but part of the key material
+				// is overwrite, so we don't fully compare the 32 bytes.
```
arkpar, line 263: Value table always contains 26 bytes of the key. The rest are checked in the index. The index start with 16 bits of chunk index and 38 bit of key material, so 54 key bits total. After the first reindex, chunk index increases to 17 bit and the key material decreases to 37, but total is still 54. And so on.

src/index.rs
cheme: Yes I don't remember very well, was probably thinking about a db directory abstraction, but it is more an idea for code organisation. No conflict with current code can happen indeed.

src/table.rs
cheme: :facepalm: works indeed.
I keep liking the idea to store the full hashed key, because it would allow restoring/rebuilding indexes from value table only, not that interesting though.

src/column.rs
arkpar: > So even if all reindexing run before this log enactment, the size will only be half.

Unless after the reindex, all the items from the chunk still end up in a single chunk, as they all share the same additional bit. Which, I agree, is extremely unlikely.

src/db.rs
arkpar: OK, I guess it can be renamed

src/column.rs
cheme: Oh yes, I did not think that way. So it is 'just' if the 64 entry got the same bit at the n + 1 positon, unlikely, but better checked :+1:

arkpar:
Most of it has been addressed I believe. If there's still anything worth further discussion please open a separate issue.

PR #23, opened 2021-04-30
BE reindex

cheme:
Quick fix for index transmute in be arch, and few utility function for index.

Removing the test and the .sh before merge could be a good idea (not that useful, except for implementing).


arkpar:
Looks good but needs a rebase/merge

cheme:
Merged master (and remove sh).

PR #28, opened 2021-06-22
Make sure data files are flushed in sync mode

arkpar:


cheme (approved):
LGTM, sync seem to have a high cost though, but if synching doing it like this seems strictly better.

Maybe there can be a way to keep flushing file a bit longer alive to reduce number of flush, but no point in thinking about it at this point.

src/log.rs
```diff
@@ -521,6 +521,8 @@ impl Log {
 				log::debug!(target: "parity-db", "Flush: Awaiting log reader");
 				self.done_reading_cv.wait(&mut reading_state)
 			}
+			// Call reader callback
+			on_read_complete()?;
```
cheme, line 525:
```suggestion
			if self.sync {
				on_read_complete()?;
			}
```

PR #30, opened 2021-07-13
Table relative indexes read and write.

cheme:
This PR uses read and write with offset api on table.

Not sure it is very interesting, it was part of a different branch with different table layout, and I was wondering about removing Rc when not needed, and it may be easier with this code as a start.

### DIFF
--- a/src/table.rs
+++ b/src/table.rs
@@ -85,7 +85,7 @@ const LOCKED_REF: u32 = 0x7fff_ffff;
 pub type Key = [u8; KEY_LEN];
 pub type Value = Vec<u8>;
 
-fn table_slice(hash: &Key) -> &[u8] {
+fn partial_key(hash: &Key) -> &[u8] {
 	&hash[6..]
 }
 
@@ -176,7 +176,7 @@ type PartialKeyEntry = Entry<[u8; 40]>;
 
 impl<B: AsRef<[u8]> + AsMut<[u8]>> Entry<B> {
 	#[inline(always)]
-	fn new() -> Self {
+	fn new_uninit() -> Self {
 		Entry(0, unsafe { MaybeUninit::uninit().assume_init() })
 	}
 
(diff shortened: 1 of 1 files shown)


src/table.rs
```diff
@@ -143,6 +151,107 @@ fn disable_read_ahead(_file: &std::fs::File) -> Result<()> {
...
+}
+
+struct Entry<B: AsRef<[u8]> + AsMut<[u8]>>(usize, B);
+type FullEntry = Entry<[u8; MAX_ENTRY_SIZE]>;
+type PartialEntry = Entry<[u8; 10]>;
+type PartialKeyEntry = Entry<[u8; 40]>;
+
+impl<B: AsRef<[u8]> + AsMut<[u8]>> Entry<B> {
+	#[inline(always)]
+	fn new() -> Self {
```
arkpar, line 179: Should be called `new_uninit` to make it explicit that it is not initialized.

src/table.rs
```diff
@@ -234,54 +343,58 @@ impl ValueTable {
...
-		let mut buf: [u8; MAX_ENTRY_SIZE] = unsafe { MaybeUninit::uninit().assume_init() };
+		let mut buf = FullEntry::new();
 
 		let mut part = 0;
 		let mut compressed = false;
 		let entry_size = self.entry_size as usize;
 		loop {
-			let buf = if log.value(self.id, index, &mut buf) {
-				&buf
+			let buf = if log.value(self.id, index, &mut buf.1) {
```
arkpar, line 352: There's a lot of `buf.1` in the code now, which makes it a bit harder to read. Looks like `buf.as_mut()` should work here because of `AsMut`. Let's use it instead.

src/table.rs
```diff
@@ -534,26 +646,29 @@ impl ValueTable {
...
 			&mut buf
 		} else {
-			self.read_at(&mut buf[0..self.entry_size as usize], index * self.entry_size as u64)?;
-			&mut buf[0..self.entry_size as usize]
+			self.read_at(&mut buf.1[..self.entry_size as usize], index * self.entry_size as u64)?;
+			&mut buf
 		};
 
-		if &buf[0..2] == TOMBSTONE {
+		let size = buf.read_size();
```
arkpar, line 657: Looks like this can be done after the `is_tombstone` check below.

src/table.rs
```diff
@@ -234,54 +343,58 @@ impl ValueTable {
...
+			let size = buf.read_size();
+
+			if buf.is_tombstone() {
 				return Ok((false, false));
 			}
 
-			let (content_offset, content_len, next) = if &buf[0..2] == MULTIPART {
-				let next = u64::from_le_bytes(buf[2..10].try_into().unwrap());
-				(10, entry_size - 10, next)
+			let (end_entry, next) = if buf.is_multipart() {
```
arkpar, line 372: `end_of_entry` or `entry_end`

src/table.rs
```diff
@@ -81,6 +85,10 @@ const LOCKED_REF: u32 = 0x7fff_ffff;
 pub type Key = [u8; KEY_LEN];
 pub type Value = Vec<u8>;
 
+fn table_slice(hash: &Key) -> &[u8] {
```
arkpar, line 88: Let's rename to `partial_key` for consistency

PR #31, opened 2021-07-13
Only exit commit after log last flush.

cheme:
When using the stress exec, I had some failure shutdown, this should fix it. (currently if commit thread exit before flushing one, the flushing cannot perform it last action)


cheme:
Useless with #32

PR #33, opened 2021-07-16
Remove REF from value table when unneeded.

cheme:
This PR remove Rc from value when not needed.
When rc is not use, but compression is possible (active in option and on a table above compression threshold), then a bit is added to indicate if compression is used.
Also rc uses full 32 byte when no compression is possible.

### DIFF
--- a/src/table.rs
+++ b/src/table.rs
@@ -23,10 +23,10 @@
 //
 // Complete entry:
 // [SIZE: 2][REFS: 4][KEY: 26][VALUE: SIZE - 30]
-// SIZE: 16-bit value size. Sizes up to 0xfffd are allowed.
-// This includes size of REFS and KEY
-// REF: 31-bit reference counter, first bit reserved to flag an applied compression
-// when needed (32-bit counter otherwhise).
+// SIZE: 15-bit value size. Sizes up to 0x7ffd are allowed.
+// This includes size of REFS and KEY.
+// The first bit is reserved to indicate if compression is applied.
+// REF: 32-bit reference counter.
 // If collection is not ref counted, a single bit is used to indicate
 // if compression is needed.
 // this is removed or replaced by one byte for applied compression when need.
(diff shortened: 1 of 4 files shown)


src/table.rs
```diff
@@ -25,7 +25,9 @@
 // [SIZE: 2][REFS: 4][KEY: 26][VALUE: SIZE - 30]
 // SIZE: 16-bit value size. Sizes up to 0xfffd are allowed.
 // This includes size of REFS and KEY
-// REF: 31-bit reference counter, first bit reserved to flag an applied compression.
+// REF: 31-bit reference counter, first bit reserved to flag an applied compression
+// when needed (32-bit counter otherwhise). If collection is not ref counted,
+// this is removed or replaced by one byte for applied compression when need.
```
cheme, line 30:
```suggestion
// this is removed or replaced by one byte to apply compression when needed.
```

arkpar:
Let's keep the compression bit regardless of the compression threshold. So that it could be changed in the options and not require a database migration.

arkpar:
We could also store the compression bit in the `SIZE` field rather than `REF`. Just need to limit the entry size to 0x7ffd rather than 0xfffd

cheme:
> 0x7ffd

Sounds like a good idea, yet 0x7ffd is a bit bellow the current max size default (32768 0x8000).
But could also add a byte in case the size tier requires the full 2 bytes (saving on header size only make sense for small value).
Could also make size = previous_threshold size + written_size.
Or just reduce max default size.

arkpar:
Reducing max default size is the most simple thing to do.

cheme:
Seems better, the only drawback is that we need to read the whole multipart to know if it is compress (but with current api it makes no difference).
Also  https://github.com/paritytech/parity-db/pull/33/commits/6881785e97dc52ea910ee59af61f3f55d1ab78ff contains bug fix that also exists in master.

arkpar (approved):

PR #34, opened 2021-07-19
Migration utility and other fixes

arkpar:
Migration allows changing column settings, such as compression and value tiers, for an existing database.


bench/Cargo.toml
```diff
@@ -1,17 +1,21 @@
 [package]
-name = "db-bench"
+name = "stress"
```
arkpar, line 2: A bit of git screwup. This should be reviewed as a new file, not a rename. `bench/core/` does not exist anymore.

migrate/Cargo.toml
```diff
@@ -1,13 +1,18 @@
 [package]
-name = "stress"
+name = "migrate"
```
arkpar, line 2: A bit of git screwup. This should be reviewed as a new file, not a rename. `bench/core` does not exist anymore.

src/column.rs
```diff
@@ -29,7 +29,7 @@ use crate::{
 use crate::compress::Compress;
 
 const START_BITS: u8 = 16;
-const MAX_REBALANCE_BATCH: u32 = 1024;
+const MAX_REBALANCE_BATCH: usize = 8192;
```
arkpar, line 32: This is now counted in entries, rather than chunks

src/log.rs
```diff
@@ -368,7 +368,6 @@ struct Reading {
 enum ReadingState {
 	Reading,
 	Idle,
-	Shutdown,
```
arkpar, line 371: `Shutdown` is not actually needed. Log worker thread makes sure that the log is read till the end before exiting, so we don't leave the reader halfway done.

src/options.rs
```diff
@@ -85,6 +91,34 @@ impl ColumnOptions {
 		}
 		true
 	}
+
+	fn from_string(s: &str) -> Option<Self> {
```
cheme, line 95: This function is not ideal to write/maintain, but having a json lib deps does not seems good either.
Maybe switching simple .properties file at some point could be good.

src/migration.rs
```diff
@@ -0,0 +1,126 @@
...
+	// Make sure we are using the same salt value.
+	to.salt = salt;
+
+	let source = Db::open(&source_options)?;
+	let dest = Db::open(&to)?;
+
+	let mut ncommits: u64 = 0;
+	let mut commit = Vec::with_capacity(COMMIT_SIZE);
+	let mut last_time = std::time::Instant::now();
+	for c in 0 .. source_options.columns.len() as ColId {
```
cheme, line 45: Migration could be doable on a per-column basis.

src/options.rs
```diff
@@ -122,18 +157,42 @@ impl Options {
...
+			for c in 0..cols.len() {
+				cols[c].compression_treshold = self.columns[c].compression_treshold;
+				if cols[c] != self.columns[c] {
+					return Err(Error::InvalidConfiguration(format!(
+								"Column config mismatch for column {}. Expected \"{}\", got \"{}\"",
+								c, self.columns[c].as_string(), cols[c].as_string())));
+				}
+			}
+		} else {
+			let s: Salt = self.salt.unwrap_or(rand::thread_rng().gen());
```
cheme, line 178: So with these changes we always be using salt?

src/options.rs
```diff
@@ -145,25 +204,26 @@ impl Options {
 					return Err(Error::InvalidConfiguration(format!(
 						"Unsupported database version {}. Expected {}", version, CURRENT_VERSION)));
 				}
+				if version == 3 {
+					//Treat all tables as ref counted.
+					force_rc = true;
```
cheme, line 209: I am not sure the previous change in rc will migrate successfully.
There can be entry with size above the new default max default table size for whom we will read an incorrect size.

src/table.rs
```diff
@@ -374,18 +390,20 @@ impl ValueTable {
 		Ok(())
 	}
 
-	// Return if there was content, and if it was compressed.
+	// Return ref counter, and if it was compressed.
```
cheme, line 393:
```suggestion
	// Return ref counter, partial key and if it was compressed.
```

src/table.rs
```diff
@@ -22,23 +22,20 @@
...
+// REF: 32-bit reference counter (optional).
 // KEY: lower 26 bytes of the key.
-// VALUE: SIZE-30  payload bytes.
+// VALUE: payload bytes.
 //
 // Partial entry (first part):
-// [MULTIPART: 2][NEXT: 8][REFS: 4][KEY: 26][VALUE]
-// MULTIPART - Split entry marker. 0xfffe.
+// [MULTIHEAD: 2][NEXT: 8][REFS: 4][KEY: 26][VALUE]
+// MULTIHEAD - Split entry head marker. 0xfffd.
```
cheme, line 35: So we flag head of multipart, not sure why it is needed (for value iteration maybe, but I did  see it only for preimage and it looked like we where using the indexing for the multiparts)?

cheme (commented):
I like the idea to use a dest_file instead of complex commands parameters to migrate.
I also did use similar admin tools in my working compression branch, so I am wondering if having a 'admin' binary instead of 'migrate' with additional commands from this branch (thinking 'clear_stats', 'flush_logs', and 'stress') could be interesting. (for it I did add struct_opt as dep, but can be just a dep of the admin crate).

cheme (approved):
Changes lgtm, added a few comment.

src/migration.rs
cheme: (just realized it won't be very user friendly)

src/column.rs
arkpar: Depends on how full the chunks would be.  The problem with the previous chunk-based approach is that when reindexing a mostly-empty index it would produce empty batches and reindexing would stop.

src/column.rs
```diff
@@ -513,6 +513,66 @@ impl Column {
...
 	}
 
+	pub fn iter_while(&self, log: &Log, mut f: impl FnMut (u64, Key, u32, Vec<u8>) -> bool) -> Result<()> {
+		let tables = self.tables.read();
+		let source = &tables.index;
+
+		if self.preimage {
+			// It is much faster to iterate over the value table than index.
+			// We have to assume hashing scheme however.
+			for table in &tables.value[..tables.value.len() - 1] {
```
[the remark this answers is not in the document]
arkpar: `tables.value[..15]` this range does not actually include index 15

src/db.rs
```diff
@@ -215,30 +215,34 @@ impl DbInner {
...
+	}
+
+	fn commit_raw(&self, commit: Vec<(ColId, Key, Option<Value>)>) -> Result<()> {
 		{
 			let mut queue = self.commit_queue.lock();
 			if queue.bytes > MAX_COMMIT_QUEUE_BYTES {
 				log::debug!(target: "parity-db", "Waiting, qb={}", queue.bytes);
 				self.commit_queue_full_cv.wait(&mut queue);
 			}
+			{
```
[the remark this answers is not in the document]
arkpar: There was a deadlock possible when the background thread stops because of the error. Commit would wait here indefinitely. Now the error wakes up the commiting thread and the error is checked after that.

src/db.rs
```diff
@@ -215,30 +215,34 @@ impl DbInner {
...
-		}
-
 		let commit: Vec<_> = tx.into_iter().map(
 			|(c, k, v)| (c, self.columns[c as usize].hash(k.as_ref()), v)
 		).collect();
 
+		self.commit_raw(commit)
+	}
+
+	fn commit_raw(&self, commit: Vec<(ColId, Key, Option<Value>)>) -> Result<()> {
```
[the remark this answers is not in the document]
arkpar: It could be done in the log writer thread. We want to make as little work in the `commit` function as possible, so that it returns as early as possible. 
I was considering changing the API from the `(column, key, Option<value>)` iterator to more of `(column, Action)` where `Action` is one of `Insert`, `Remove`, `ChangeRef`. We will need something like this for subindexes anyway.

src/options.rs
arkpar: This just allows overriding the salt in the options, but does not change the logic. The salt won't be set for an existing metadata that does not already has it.

src/column.rs
```diff
@@ -347,7 +347,7 @@ impl Column {
 				}
 				if self.ref_counted {
 					log::trace!(target: "parity-db", "{}: Increment ref {}", tables.index.id, hex(key));
-					tables.value[target_tier].write_inc_ref(existing_address.offset(), log, compressed)?;
+					tables.value[existing_tier].write_inc_ref(existing_address.offset(), log, compressed)?;
```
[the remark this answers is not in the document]
arkpar: Refactored the code a bit to avoid id.

src/migration.rs
```diff
@@ -0,0 +1,126 @@
...
+		log::info!("Migrating col {}", c);
+		source.iter_column_while(c, |index, key, rc, mut value| {
+			//TODO: more efficient ref migration
```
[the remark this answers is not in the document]
arkpar: I went for a more general approach, so that the values could be migrated between value tables, which requires index update. This seems like the simplest way to do it for now.

src/options.rs
arkpar: These should produce a decompression error, which I this is fine for now. I guess we could additionally check that compression bit is not set for columns with no `compression` option as well

src/table.rs
arkpar: I thought I left a comment about this, but it got lost. Yes, this is just so that the value iterator could be implemented in the future, once the new format is adopted.

arkpar:
+1 For having a generic admin tool

src/options.rs
cheme: Only if there is a compression (the bit is also used for NoCompression to allow switching). Additional check seems worth it.

cheme:
> +1 For having a generic admin tool

will do a PR for it this week

src/options.rs
arkpar: Will add in the next PR

PR #35, opened 2021-07-20
Introduce metadata struct & compatibility

arkpar:


src/db.rs
```diff
@@ -96,6 +96,7 @@ impl std::hash::Hasher for IdentityKeyHash {
 struct DbInner {
 	columns: Vec<Column>,
 	options: Options,
+	metadata: Metadata,
```
cheme, line 99: ColumnOptions being stored in both metadata and options, does not look problematic.

src/options.rs
```diff
@@ -23,7 +23,7 @@ use crate::column::Salt;
 use crate::compress::CompressionType;
 use rand::Rng;
 
-const CURRENT_VERSION: u32 = 4;
+pub const CURRENT_VERSION: u32 = 4;
 const LAST_SUPPORTED_VERSION: u32 = 3;
```
cheme, line 27:
```suggestion
// TODO on last supported 4, remove `ValueTable` `no_compression` field.
const LAST_SUPPORTED_VERSION: u32 = 3;
```
or similar comment to not forget it.

cheme (approved):
LGTM

src/db.rs
arkpar: Semantically columns settings in options have a meaning of what the database metadata should be when it is created anew. Might be a good idea to split `Db::open` into `create` that takes column settings and `open` that does not.

PR #36, opened 2021-07-22
Admin commands.

cheme:
This PR groups all external tools in a 'parity-admin-crate'.
I added commands I was using while working on compression. Not all may be useful so please chose which to remove.

I also change a bit the migration to avoid useless work and allow compressing only a single column (use the existing meta file edited with a change compression in column and use the --overwrite flag).


admin/src/bench/db.rs
```diff
@@ -14,11 +14,13 @@
 // You should have received a copy of the GNU General Public License
 // along with Parity.  If not, see <http://www.gnu.org/licenses/>.
 
-pub type Key = [u8; 32];
-pub type Value = Vec<u8>;
-
+use parity_db::{Key, Value};
+// TODO useless??
```
cheme, line 18: The bench could probably be change to remove this trait, I am keeping it in this PR to avoid changing code.

admin/src/bench/mod.rs
```diff
@@ -30,40 +32,97 @@ static COMMITS: AtomicUsize = AtomicUsize::new(0);
...
+pub(super) struct BenchAdapter(parity_db::Db);
+
+impl BenchDb for BenchAdapter {
+	type Options = parity_db::Options;
+
+	fn open(path: &std::path::Path) -> Self {
+		BenchAdapter(Db::with_columns(path, 1).unwrap())
+	}
+
+	fn with_options(options: &Self::Options) -> Self {
```
cheme, line 50: Only this function was added over existing code. (the diff is a bit bad here).

admin/src/bench/mod.rs
```diff
@@ -30,40 +32,97 @@ static COMMITS: AtomicUsize = AtomicUsize::new(0);
...
-			append: false,
-			seed: None,
-			archive: false,
+			readers: self.readers.unwrap_or(4),
+			writers: self.writers.unwrap_or(1),
+			commits: self.commits.unwrap_or(100_000),
+			seed: self.seed.clone(),
+			append: self.append,
+			archive: self.archive,
+			check_only: self.check_only,
```
cheme, line 124: actually check_only and start_commit looks new

admin/src/bench/db.rs
arkpar: It was introduced to be able to plugin in alternative databases and compare performance. Subdb was supported initially, but that was later dropped.

admin/src/bench/mod.rs
```diff
@@ -30,40 +32,97 @@ static COMMITS: AtomicUsize = AtomicUsize::new(0);
...
+	#[structopt(long)]
+	pub append: bool,
+
+	/// Do not apply pruning.
+	#[structopt(long)]
+	pub archive: bool,
+
+	/// Do only check (requires append flag).
+	#[structopt(long)]
+	pub check_only: bool,
```
arkpar, line 100: Do we actually need that? `--readers=0 --writers=0` has the same effect, doesn't it? I'd rather have `--no-check` that skips final verification.

admin/src/bench/mod.rs
arkpar: I was thinking we should actually put the last commit number in the database along with key values. Also put the seed in the first commit. This would allow for correct continuation with `--append`

admin/src/bench/mod.rs
```diff
@@ -239,50 +252,57 @@ pub fn run<D: Db>() {
...
-	};
+	println!(
+		"Completed {} commits in {} seconds. {} cps",
+		commits,
+		elapsed,
+		commits as f64  / elapsed
+	);
 
-	let db = Arc::new(Db::open(path.as_path())) as Arc<D>;
+	thread::sleep(std::time::Duration::from_secs(1));
```
arkpar, line 276: Database is now properly flushed on shutdown, so this may now be removed.

admin/src/lib.rs
```diff
@@ -0,0 +1,323 @@
...
+				check.display,
+				check.display_value_max,
+			);
+			db.check_from_index(check_param)
+				.map_err(|e| format!("Check error: {:?}", e))?;
+		},
+		SubCommand::Flush(_flush) => {
+			let db = parity_db::Db::open(&options)
+				.map_err(|e| format!("Invalid db: {:?}", e))?;
+			while !db.all_empty_logs() {
```
arkpar, line 126: This should really be just `{ parity_db::Db::open() }`. It is flushed when dropped.

src/column.rs
```diff
@@ -514,13 +514,30 @@ impl Column {
...
+		let tables = self.tables.read();
+		let empty_stats = ColumnStats::empty();
+		tables.index.write_stats(&empty_stats);
+	}
+
+	pub fn iter_while(&self, log: &Log, f: impl FnMut (u64, Key, u32, Vec<u8>) -> bool) -> Result<()> {
+		self.iter_while_inner(log, f, 0,  None, |_, _| false)
+	}
+
+	fn iter_while_inner(
```
arkpar, line 533: Feels like this function is now doing too much. Maybe let's keep a separate one for checking consistency. Also you don't need start and end chunk when you have the callback. 
So it should either be  `fn check_consistency(&self, log: &Log, start_chunk: u64, sub_index: u8, end_chunk: Option<u64>)-> Result<(), ConsistencyError>` which returns on the first error. `ConsistencyError` contains chunk, entry with sub index so that you can continue checking from the next entry.

Or it could be `check_consistency(&self, log: &Log, f: impl FnMut (Result<Progress, ConsistencyError>) -> bool`
which calls `f` for each with progress or error information and decides if to stop based on what the callback returns.

arkpar: Also, consistency check should not use preimage iteration over the value table, as the index is not really validated in this mode.

src/db.rs
```diff
@@ -655,16 +657,35 @@ impl DbInner {
...
+					let mut writer = std::io::BufWriter::new(file);
+					self.do_collect_stats(&mut writer, None)
 				}
 				Err(e) => log::warn!(target: "parity-db", "Error creating stats file: {:?}", e),
 			}
 		}
 		Ok(())
 	}
 
+	fn do_collect_stats(&self, writer: &mut impl std::io::Write, column: Option<u8>) {
```
arkpar, line 669: Let's drop the `do_` prefix here and in `do_clear_stats` if possible

src/db.rs
```diff
@@ -826,6 +869,29 @@ impl Db {
...
+
+	pub fn do_collect_stats(&self, writer: &mut impl std::io::Write, column: Option<u8>) {
+		self.inner.do_collect_stats(writer, column)
+	}
+
+	pub fn do_clear_stats(&self, column: Option<u8>) {
+		self.inner.do_clear_stats(column)
+	}
+
+	pub fn all_empty_logs(&self) -> bool {
```
arkpar, line 881: This should not be needed ever

src/migration.rs
```diff
@@ -66,11 +88,63 @@ pub fn migrate(from: &Path, mut to: Options) -> Result<()> {
...
+			std::mem::drop(dest);
+			dest = Db::open_or_create(&to)?;
+			while !dest.all_empty_logs() {
+				std::thread::sleep(std::time::Duration::from_millis(300));
+			}
+			log::info!("Collection migrated {}, imported", c);
+
+			std::mem::drop(dest);
+			std::mem::drop(source);
+			move_collection(c, &to.path, from)?;
```
arkpar, line 103: MIgration should not touch source. Everything should be copied for case of failures we don't want to end up with both the source and dest ruined. During the upgrade only after we have migrated the whole database successfully we can switch to it and delete the old one.

src/migration.rs
```diff
@@ -21,29 +21,51 @@ use crate::{options::Options, db::Db, Error, Result, column::ColId};
 
 const COMMIT_SIZE: usize = 10240;
 
-pub fn migrate(from: &Path, mut to: Options) -> Result<()> {
-	let mut path: std::path::PathBuf = from.into();
-	path.push("metadata");
-	let source_meta = Options::load_metadata(&path)?
+pub fn migrate(from: &Path, mut to: Options, keep_dest: bool, force_migrate: &Vec<u8>) -> Result<()> {
```
arkpar, line 24: What is the `keep_dest` option for?

src/migration.rs
```diff
@@ -66,11 +88,63 @@ pub fn migrate(from: &Path, mut to: Options) -> Result<()> {
...
 
+fn move_collection(c: ColId, from: &Path, to: &Path) -> Result<()> {
+	deplace_collection(c, from, to, false)
+}
+
+fn copy_collection(c: ColId, from: &Path, to: &Path) -> Result<()> {
+	deplace_collection(c, from, to, true)
+}
+
+fn deplace_collection(c: ColId, from: &Path, to: &Path, copy: bool) -> Result<()> {
```
arkpar, line 126: Should be called `copy_column`

admin/src/lib.rs
```diff
@@ -0,0 +1,323 @@
...
+	/// Only show stat for the given column.
+	#[structopt(long)]
+	pub column: Option<u8>,
+
+	/// Clear current stats.
+	#[structopt(long)]
+	pub clear: bool,
+}
+
+/// Migrate db (update version or change collection options).
```
arkpar, line 251: Using multiple terms for the same thing is a source of confusion. Let's stick to "column"

admin/src/bench/mod.rs
```diff
@@ -30,40 +32,97 @@ static COMMITS: AtomicUsize = AtomicUsize::new(0);
 //static QUERIES: AtomicUsize = AtomicUsize::new(0);
 
 const COMMIT_SIZE: usize = 100;
+// Note on db, this is equal to COMMIT_SIZE.
+// Having smaller allow validating a given amount
+// of keys.
```
arkpar, line 37:
```suggestion
// Out of `COMMIT_SIZE` values `COMMIT_PRUNE_SIZE` will be deleted in a later commit.
// The rest will be queried during the final check.
```

admin/src/lib.rs
```diff
@@ -0,0 +1,323 @@
...
+// Note standard substrate db uses:
+// NUM_COLUMN 12
+// Column 1 (state):
+// 
+//		state_col.ref_counted = true;
+//		state_col.preimage = true;
+//		state_col.uniform = true;
+//
+// TODO Passing parameter to define this is be a bit tedious,
+// using a existing metadata file could be good.
```
arkpar, line 28: What exactly does this TODO suggest? Note that column settings for substrate are likely to change soon. And it should not matter for this tool anyway.

admin/src/lib.rs
```diff
@@ -0,0 +1,323 @@
...
+	pub no_sync: bool,
+
+	/// Register stat from those admin operations.
+	#[structopt(long)]
+	pub with_stats: bool,
+
+	/// Indicate the number of column, when using
+	/// a new or temporary db, defaults to one.
+	#[structopt(long)]
+	pub nb_columns: Option<u8>,
```
arkpar, line 167: Could be simply `columns`

admin/src/lib.rs
```diff
@@ -0,0 +1,323 @@
...
+
+	/// Indicate a metadata configuration to use.
+	#[structopt(long)]
+	pub use_metadata: Option<PathBuf>,
```
arkpar, line 171: This only makes sense for the migration. All the commands that require an existing database should simply accept path.

admin/src/lib.rs
```diff
@@ -0,0 +1,323 @@
...
+
+	/// Destination or temporary folder for migration.
+	#[structopt(long)]
+	pub dest_path: PathBuf,
+
+	/// Meta to migrate to.
+	#[structopt(long)]
+	pub dest_meta: PathBuf,
+
+	/// Destina
```
arkpar, line 279:
```suggestion
	/// Global compression threshold for columns with compression enabled.
```

src/db.rs
```diff
@@ -694,17 +715,39 @@ impl Db {
...
+	/// Create the database using given options.
+	pub fn open_or_create(options: &Options) -> Result<Db> {
+		Self::open_inner(options, true, false)
+	}
+
+	pub fn open_read_only(options: &Options) -> Result<Db> {
+		Self::open_inner(options, false, true)
+	}
+
+	pub fn open_inner(options: &Options, create: bool, no_thread: bool) -> Result<Db> {
```
arkpar, line 735:
```suggestion
	pub fn open_inner(options: &Options, create: bool, read_only: bool) -> Result<Db> {
```

src/db.rs
```diff
@@ -694,17 +715,39 @@ impl Db {
...
+		assert!(options.is_valid());
+		let mut db = DbInner::open(options, create)?;
 		// This needs to be call before log thread: so first reindexing
 		// will run in correct state.
 		db.replay_all_logs()?;
 		let db = Arc::new(db);
+		if no_thread {
```
arkpar, line 742:
```suggestion
		if read_only {
```

src/db.rs
```diff
@@ -840,3 +906,32 @@ impl Drop for Db {
 		}
 	}
 }
+
+/// Verification operation utilities.
+pub mod check {
+	pub struct CheckParam {
```
arkpar, line 912: Let's call this `CheckOptions` since we already use "options" for the database.

src/db.rs
```diff
@@ -840,3 +906,32 @@ impl Drop for Db {
...
+		pub column: Option<u8>,
+		pub from: Option<u64>,
+		pub bound: Option<u64>,
+		pub display_content: bool,
```
arkpar, line 916: could be an `enum { None, Full, Short(u64) }`

src/options.rs
```diff
@@ -252,3 +252,15 @@ impl Options {
 		true
 	}
 }
+
+impl Metadata {
+	pub fn column_to_migrate(&self) -> std::collections::BTreeSet<u8> {
```
arkpar, line 257:
```suggestion
	pub fn columns_to_migrate(&self) -> std::collections::BTreeSet<u8> {
```

arkpar (changes requested):
Great work, but requires some minor changes.

admin/src/bench/mod.rs
cheme: readers 0 and writers 0 seems to stall, I would not really thought of using this combination, so I guess keeping the parameter can make sense. +1 for --no-check.

cheme: Actually writers 0 could replace check only and readers 0 could replace no-check

cheme: well check_only without the start parameter is useless, removing it.

arkpar: Rather `--commit=0` should have the same effect

src/column.rs
cheme: Yes 'end_chunk' does not make sense.
I did plan to suspend/restore iteration on error at first, but it would need to manage sub_index when starting, using a call back was a bit easier. So will use an enum parameter on the callback instead.

src/migration.rs
cheme: My use case is changing column option that are not strict (eg compression or the size threshold) for a column.
So in this case I am not really 'migrating' but simply rewriting a collection ('keep_dest' set to false or 'overwrite' parameter to true).
When the metadata version number is increased, this is not possible (there is a check above).

src/migration.rs
cheme: I will rename it to overwrite, it changes a bit the semantic of migration.

admin/src/lib.rs
cheme: I had previously everything as parameter, so when I wanted to change some metadata I had to redefine all compression for all columns.
So the suggestion was to write a reading metadata method to update it from parameter (which is in you prev pr).
But now we use a dest_metadata so this is totally out of date.

admin/src/lib.rs
cheme: For migration, is there a case where we want to override the source metadata?
I will just remove it.

src/db.rs
```diff
@@ -120,11 +120,13 @@ struct DbInner {
...
-	fn open(options: &Options) -> Result<DbInner> {
-		std::fs::create_dir_all(&options.path)?;
+	fn open(options: &Options, create: bool) -> Result<DbInner> {
+		if create {
+			std::fs::create_dir_all(&options.path)?
+		};
 		let mut lock_path: std::path::PathBuf = options.path.clone();
 		lock_path.push("lock");
-		let lock_file = std::fs::OpenOptions::new().create(true).read(true).write(true).open(lock_path.as_path())?;
+		let lock_file = std::fs::OpenOptions::new().create(create).read(true).write(true).open(lock_path.as_path())?;
```
arkpar, line 129: Creation of lock file should be attempted regardless of the `create` flag. I.e. when opening an existing database with no lock, it should be created and taken.

admin/src/lib.rs
arkpar: I can't think of such case. Metadata should always match source tables.

src/migration.rs
```diff
@@ -66,11 +88,60 @@ pub fn migrate(from: &Path, mut to: Options) -> Result<()> {
...
+				|| crate::table::TableId::is_file_name(c, file) {
+
+				let mut from = from.to_path_buf();
+				from.push(file);
+				let mut to = to.to_path_buf();
+				to.push(file);
+				if copy {
+					std::fs::copy(from, to)?;
+				} else {
+					std::fs::rename(from, to)?;
```
arkpar, line 137: After #37 source and dest might end up with a different set of tables. So after "rename" some junk tables may be left.

src/db.rs
cheme: :+1: (the locks usually exists but being able to remove such file is something expected)

src/migration.rs
cheme: :+1: 
Looking at files dates, I did already have one.

PR #37, opened 2021-07-22
Move from 16 to 256 value tables

arkpar:
The original plan was to have 16 size tiers and allow users to tweak them so that size amplification is minimized. In practice though it requires arcane knowledge to do so and the optimal values are likely to become less optimal over time as more stuff is added to the database. It makes more sense to opt for a more simple configuration.

The implementation is changed to crate the value tables only when they are needed.

The downside is degraded write performance on MacOS 11. There seems to be an issue with the page caching there. Luckily it seems to fixed in MacOS 12 beta. Linux does not seem affected much. Another downside is that address space limit is decreased by 4 bits. 

Polkadot database size is reduced by about 30% when migrated to a new format.


src/db.rs
```diff
@@ -594,7 +594,7 @@ impl DbInner {
 	}
 
 	fn cleanup_logs(&self) -> Result<bool> {
-		let keep_logs = if self.options.sync_data { KEEP_LOGS } else { 0 };
+		let keep_logs = if self.options.sync_data { 0 } else { KEEP_LOGS };
```
arkpar, line 597: This was a bug. The values should be reversed. We only need to keep `KEEP_LOGS` old files when sync is disabled. This gives at least some chance to recover after the power failure or OS crash.

src/column.rs
```diff
@@ -581,11 +571,21 @@ impl Column {
 				if entry.is_empty() {
 					continue;
 				}
-				let address = entry.address(source.id.index_bits());
-				if skip_preimage_indexes && self.preimage && address.size_tier() != 15 {
+				let (size_tier, offset) = if self.db_version >= 4 {
+					let address = entry.address(source.id.index_bits());
+					(address.size_tier(), address.offset())
+				} else {
+					let addr_bits = source.id.index_bits() + 10;
```
cheme, line 578:
```suggestion
					let addr_bits = source.id.index_bits() + CHUNK_ENTRIES_BITS + SIZE_TIERS_BITS - 4;

```

arkpar: This should not be defined in terms of current value of `SIZE_TIERS_BITS`. If `SIZE_TIERS_BITS` changes in the future, this would need to be updated too. I think a hardcoded value is fine here. V3 had exactly  + 10 bits and that's not going to change.

src/column.rs
```diff
@@ -383,14 +370,15 @@ impl Column {
...
-				log::trace!(target: "parity-db", "{}: Inserting new index {}", tables.index.id, hex(key));
 				let (cval, target_tier) = self.compress(&key, &val, &*tables);
 				let (cval, compressed) = cval.as_ref()
 					.map(|cval| (cval.as_slice(), true))
 					.unwrap_or((val.as_slice(), false));
 
+				log::trace!(target: "parity-db", "{}: Inserting new index {}, size = {}", tables.index.id, hex(key), cval.len());
 				let offset = tables.value[target_tier].write_insert_plan(key, &cval, log, compressed)?;
 				let address = Address::new(offset, target_tier as u8);
+				assert!(address.size_tier() == target_tier as u8);
```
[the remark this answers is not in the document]
arkpar: Just made sure `Address` is implemented correctly. I'll remove it.

src/column.rs
```diff
@@ -581,11 +571,21 @@ impl Column {
...
+					(address.size_tier(), address.offset())
+				} else {
+					let addr_bits = source.id.index_bits() + 10;
+					let address = Address::from_u64(entry.as_u64() & ((1u64 << addr_bits) - 1));
+					let size_tier = (address.as_u64() & 0x0f) as u8;
+					let offset = address.as_u64() >> 4;
+					(size_tier, offset)
+				};
+
+				if skip_preimage_indexes && self.preimage && size_tier != 15 {
```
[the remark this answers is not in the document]
arkpar: Indeed

src/log.rs
```diff
@@ -337,15 +337,35 @@ impl<'a> LogQuery for LogWriter<'a> {
...
+    fn write_u32(&mut self, _: u32)     { unreachable!() }
+    fn write_u64(&mut self, n: u64)     { self.0 = n }
+    fn write_usize(&mut self, _: usize) { unreachable!() }
+    fn write_i8(&mut self, _: i8)       { unreachable!() }
+    fn write_i16(&mut self, _: i16)     { unreachable!() }
+    fn write_i32(&mut self, _: i32)     { unreachable!() }
+    fn write_i64(&mut self, _: i64)     { unreachable!() }
+    fn write_isize(&mut self, _: isize) { unreachable!() }
+    fn finish(&self) -> u64 { self.0 }
+}
```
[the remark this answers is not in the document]
arkpar: They do have different expectations on what's being hashed indeed. Better keep them separate.

src/table.rs
```diff
@@ -70,14 +72,16 @@ use crate::{
 };
 
 pub const KEY_LEN: usize = 32;
-pub const SIZE_TIERS: usize = 16;
-pub const SIZE_TIERS_BITS: u8 = 4;
+pub const SIZE_TIERS: usize = 1usize << SIZE_TIERS_BITS;
+pub const SIZE_TIERS_BITS: u8 = 8;
 pub const COMPRESSED_MASK: u16 = 0x80_00;
-const MAX_ENTRY_SIZE: usize = 0xfffc;
+pub const MAX_ENTRY_SIZE: usize = 0x7ff8;
```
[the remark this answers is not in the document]
arkpar: Additional 5 values are reserved in case we want to add additional record types in the future.

src/table.rs
```diff
@@ -319,42 +335,53 @@ impl ValueTable {
...
-		if file_len == 0 {
-			// Preallocate a single entry that contains metadata
-			file.set_len(entry_size as u64)?;
-			file_len = entry_size as u64;
+			capacity = file_len / entry_size as u64;
+			let mut header = Header::default();
+			file.read_exact(&mut header.0)?;
+			last_removed = header.last_removed();
+			filled = header.filled();
+			if filled == 0 {
```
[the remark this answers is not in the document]
arkpar: It can when file creation is interrupted with a crash. Even though it should later fixed with log replay, better be safe here.

src/options.rs
```diff
@@ -97,8 +96,11 @@ impl ColumnOptions {
 	}
 
 	pub fn is_valid(&self) -> bool {
-		for size in self.sizes {
-			if size >= crate::table::COMPRESSED_MASK {
+		if self.sizes.len() > crate::table::SIZE_TIERS {
```
cheme, line 99: I was thinking of >= (last entry for the multi entry).

cheme (approved):

src/column.rs
```diff
@@ -582,7 +581,7 @@ impl Column {
 					(size_tier, offset)
 				};
 
-				if skip_preimage_indexes && self.preimage && size_tier != 15 {
+				if skip_preimage_indexes && self.preimage && size_tier as usize != tables.value.len() - 1 {
```
cheme, line 584: Actually here also I was thinking of '!= tables.value.len()'

arkpar: `tables` here also includes the multipart table

src/options.rs
arkpar: good catch

PR #38, opened 2021-08-05
prevent Db::open from creating new database

debris:
Currently `Db::open` creates new database, even tho it shouldn't. This PR fixes this behaviour and adds 2 simple tests

### DIFF
--- a/src/db.rs
+++ b/src/db.rs
@@ -133,7 +133,7 @@ impl DbInner {
 		};
 		let mut lock_path: std::path::PathBuf = options.path.clone();
 		lock_path.push("lock");
-		let lock_file = std::fs::OpenOptions::new().create(true).read(true).write(true).open(lock_path.as_path())?;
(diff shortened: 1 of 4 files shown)


src/db.rs
```diff
@@ -133,7 +133,7 @@ impl DbInner {
 		};
 		let mut lock_path: std::path::PathBuf = options.path.clone();
 		lock_path.push("lock");
-		let lock_file = std::fs::OpenOptions::new().create(true).read(true).write(true).open(lock_path.as_path())?;
+		let lock_file = std::fs::OpenOptions::new().create(create).read(true).write(true).open(lock_path.as_path())?;
```
cheme, line 136: Initially it was done this way, but was changed to ensure we can recreate lock if it was deleted manually.
I guess directory existence check when create is false may be needed to fix this.

debris: why would someone delete it manually? without this change we have no way of telling whether parity database already exists or not

cheme (approved):
LGTM

PR #40, opened 2021-09-09
fix: madvise_random on not(unix)

kinosang:
should return () instead of Result<()>

### DIFF
--- a/src/index.rs
+++ b/src/index.rs
@@ -499,7 +499,7 @@ impl IndexTable {
 
 	#[cfg(not(unix))]
 	fn madvise_random(&self, _map: &mut memmap2::MmapMut) {
-		Ok(())
+		()
 	}
 }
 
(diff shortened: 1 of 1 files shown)


src/index.rs
```diff
@@ -499,7 +499,7 @@ impl IndexTable {
 
 	#[cfg(not(unix))]
 	fn madvise_random(&self, _map: &mut memmap2::MmapMut) {
```
shamilsan, line 501: The code might be even shorter.

```suggestion
	fn madvise_random(&self, _map: &mut memmap2::MmapMut) {}
```

src/index.rs
```diff
@@ -499,7 +499,7 @@ impl IndexTable {
...
-		Ok(())
+		()
 	}
```
shamilsan, line 503:
```suggestion
```

shamilsan:
Also waiting for this PR to be merged as it is not possible to build the project against Windows.

arkpar:
published as `0.3.2`

PR #42, opened 2021-10-31
Fix potential corruption

arkpar:
Since size is stored as 16-bit little endian in value tables, it could potentially result in a byte value that matches  `MULTIHEAD` and `MULTIPART`.  

* Changes  byte order in `MULTIHEAD` and `MULTIPART` values to avoid collisions.
* Disables multipart tracking for fixed size entry tables.
* Removed v3 compatibility

### DIFF
--- a/src/table.rs
+++ b/src/table.rs
@@ -123,10 +123,6 @@ impl TableId {
 		format!("table_{:02}_{}", self.col(), hex(&[self.size_tier()]))
 	}
 
-	pub fn legacy_file_name(&self) -> String {
-		format!("table_{:02}_{}", self.col(), self.size_tier())
-	}
-
 	pub fn is_file_name(col: ColId, name: &str) -> bool {
 		name.starts_with(&format!("table_{:02}_", col))
 	}
@@ -363,13 +359,8 @@ impl ValueTable {
 		}
 
 		let mut filepath: std::path::PathBuf = std::path::PathBuf::clone(&*path);
-		// Check for old file name format
-		filepath.push(id.legacy_file_name());
-		let mut file = if db_version == 3 && std::fs::metadata(&filepath).is_ok() {
-			Some(std::fs::OpenOptions::new().create(true).read(true).write(true).open(filepath.as_path())?)
-		} else {
-			filepath.pop();
-			filepath.push(id.file_name());
+		filepath.push(id.file_name());
+		let mut file = {
 			if std::fs::metadata(&filepath).is_ok() {
 				Some(std::fs::OpenOptions::new().create(true).read(true).write(true).open(filepath.as_path())?)
 			} else {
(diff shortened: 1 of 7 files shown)


src/table.rs
```diff
@@ -75,7 +75,7 @@ pub const KEY_LEN: usize = 32;
 pub const SIZE_TIERS: usize = 1usize << SIZE_TIERS_BITS;
 pub const SIZE_TIERS_BITS: u8 = 8;
 pub const COMPRESSED_MASK: u16 = 0x80_00;
-pub const MAX_ENTRY_SIZE: usize = 0x7ff8;
+pub const MAX_ENTRY_SIZE: usize = 0x7ff8; // Actual max size in V4 was 0x7dfe
```
cheme, line 78: Also found 0x7dfe for v4, but for new bigger entry size I am not really sure.
Should it be 7ffc: '7f' to account for compression byte and 'fc' as first value bellow any of tombstone, multipart or multihead.

src/table.rs
```diff
@@ -1254,4 +1267,24 @@ mod test {
...
+
+	#[test]
+	fn multipart_collision() {
+		let dir = TempDir::new("multipart_collision");
+		let table = dir.table(Some(super::MAX_ENTRY_SIZE as u16), &rc_options());
+		let log = dir.log();
+
+		let key = key(1);
+		let val = value(32225); // This result in 0x7dff entry size, which conflicts with v4 multipart definition
+
```
cheme, line 1279: here when testing could also use
```
		let value_size = table.value_size();
		assert_eq!(0x7fdc, table.value_size()); // Max value size for this configuration.
		let val = value(value_size as usize); // This result in 0x7ffc entry size.
```
which is more than 32225, but I am not sure if the test try to touch the biggest value.

cheme (commented):
Looks good, I am just not sure about the new value for `MAX_ENTRY_SIZE`.

arkpar:
> Looks good, I am just not sure about the new value for MAX_ENTRY_SIZE.

`MAX_ENTRY_SIZE` does not really change in this PR. I've just added a comment. Technically, now max size for fixed size tables is 0x7ffe, as they can't contain MULTIPART/MULTIHEAD records. I'd still leave it at 0x7ff8 for future proofing.

src/table.rs
arkpar: The test is checking for 32225 specifically as it result in entry size `0x7dff`, which conflicted with the previous definition of MULTIHEAD (`0xfdff`)

arkpar: I've added an additional check in the test for max value size

cheme:
> > Looks good, I am just not sure about the new value for MAX_ENTRY_SIZE.
> 
> `MAX_ENTRY_SIZE` does not really change in this PR. I've just added a comment. Technically, now max size for fixed size tables is 0x7ffe, as they can't contain MULTIPART/MULTIHEAD records. I'd still leave it at 0x7ff8 for future proofing.

cheme (approved):
In fact we could even use 0x7ff0 or a little lesser value for future use, but sounds good to me.

PR #43, opened 2021-11-08
Fixed a race condition and added a performance optimization for macos

arkpar:


cheme (approved):
Race fix lgtm, did see similar report (go) for f_fullsync with no perfect solution so let's try this.
What kind of performance difference does it make?

arkpar:
With `F_FULLSYNC` macos is about 20x slower on the stress test.  With `fsync` performance is about the same as linux. `F_FULLSYNC` used to be much faster  in MacOS Catalina, but I think in recent OS versions they have something broken in the kernel that affects performance here. I did some extensive testing and was unable to cause data corruption with fsync. This may not be the case for external drives or older HDDs, but for internal SSDs it should be fine.

athei:
> I did some extensive testing and was unable to cause data corruption with fsync. This may not be the case for external drives or older HDDs, but for internal SSDs it should be fine.

I think you only get the corruption on power loss (when the drive cache is dropped). You won't be able to test that easily on a device with a battery, right?

arkpar:
My tests involved cutting the power by long pressing the power button on a macbook.  I think this is wired in hardware to cut the power to the system regardless of the battery. Not sure what kind of power loss protection the internal SSD is using. It may be relying on system battery still or have additional capacitors. Or have none at all. Would be good to test in on a Mac mini though. I suspect that regular `flush` works for ParityDb because it still flushes everything to the NVMe controller, which creates an ordering. ParityDb can handle failed `flush` as long as write ordering is preserved. I.e. after a successful `flush` it should not be possible for writes occurred before `flush` to be lost, while writes after `flush` to be persisted.

In any case, I don't consider this to be much of an issue for substrate. Pretty much all of the validators use linux and also run in power loss protected data centres.

PR #45, opened 2021-11-15
FIx stat collection

arkpar:
Fixed incorrect accounting for compressed and uncompress bytes in the stats.

### DIFF
--- a/admin/src/bench/mod.rs
+++ b/admin/src/bench/mod.rs
@@ -22,7 +22,7 @@ use super::*;
 mod db;
 mod sizes;
 
-pub use parity_db::{Key, Value, Db};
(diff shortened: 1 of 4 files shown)


admin/src/bench/mod.rs
```diff
@@ -138,14 +150,19 @@ impl SizePool {
 		SizePool { distribution, total }
 	}
 
-	fn value(&self, seed: u64) -> Vec<u8> {
+	fn value(&self, seed: u64, compressable: bool) -> Vec<u8> {
 		let mut rng = rand::rngs::SmallRng::seed_from_u64(seed);
```
cheme, line 154: SmallRng shall compress :+1:

cheme (approved):
LGTM

PR #46, opened 2021-11-17
Extract file utility from table module.

cheme:
This PR extract file related methods from table module.
It also change the initial file capacity to align file size.

### DIFF
--- a/src/file.rs
+++ b/src/file.rs
@@ -71,44 +71,43 @@ fn fsync(file: &std::fs::File) -> Result<()> {
 }
 
 
-const GROW_SIZE: u64 = 256 * 1024;
+const GROW_SIZE_BYTES: u64 = 256 * 1024;
 
-pub struct DbFile {
+pub struct TableFile {
 	pub file: RwLock<Option<std::fs::File>>,
 	pub path: std::path::PathBuf,
 	pub capacity: AtomicU64,
 	pub dirty: AtomicBool,
-	pub origin: TableId,
+	pub id: TableId,
 }
 
-impl DbFile {
-	pub fn open(filepath: std::path::PathBuf, header_size: u8, entry_size: u16, origin: TableId) -> Result<Self> {
+impl TableFile {
+	pub fn open(filepath: std::path::PathBuf, header_entries: u8, entry_size: u16, id: TableId) -> Result<Self> {
 		let mut capacity = 0u64;
 		let file = if std::fs::metadata(&filepath).is_ok() {
 			let file = std::fs::OpenOptions::new().create(true).read(true).write(true).open(filepath.as_path())?;
 			disable_read_ahead(&file)?;
 			if file.metadata()?.len() == 0 {
 				// Preallocate.
-				debug_assert!(header_size as u64 * entry_size as u64 <= GROW_SIZE);
-				file.set_len(GROW_SIZE)?;
+				capacity += GROW_SIZE_BYTES / entry_size as u64;
+				debug_assert!(header_entries as u64 <= capacity);
+				file.set_len(capacity * entry_size as u64)?;
 			}
-			// header is part of capacity.
-			capacity = GROW_SIZE / entry_size as u64;
 			Some(file)
 		} else {
 			None
 		};
-		Ok(DbFile {
+		Ok(TableFile {
 			path: filepath,
 			file: RwLock::new(file),
 			capacity: AtomicU64::new(capacity),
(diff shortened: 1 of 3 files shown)


src/file.rs
```diff
@@ -0,0 +1,180 @@
...
+			path: filepath,
+			file: RwLock::new(file),
+			capacity: AtomicU64::new(capacity),
+			dirty: AtomicBool::new(false),
+			origin,
+		})
+	}
+
+	fn create_file(&self) -> Result<std::fs::File> {
+		log::debug!(target: "parity-db", "Created value table {}", self.origin);
```
cheme, line 111: `origin` is only use for this log.

src/file.rs
```diff
@@ -0,0 +1,180 @@
...
+		let file = std::fs::OpenOptions::new().create(true).read(true).write(true).open(self.path.as_path())?;
+		disable_read_ahead(&file)?;
```
cheme, line 113: In open file we preallocate, in create file we don't, maybe just remove preallocation or also preallocate in create file.

src/file.rs
```diff
@@ -0,0 +1,180 @@
...
+}
+
+#[cfg(not(target_os = "macos"))]
+fn fsync(file: &std::fs::File) -> Result<()> {
+	file.sync_data()?;
+	Ok(())
+}
+
+
+const GROW_SIZE: u64 = 256 * 1024;
```
arkpar, line 74: Should be `GROW_BYTES` or `GROW_SIZE_BYTES`

src/file.rs
```diff
@@ -0,0 +1,180 @@
...
+
+pub struct DbFile {
```
arkpar, line 76: Let's call it `TableFile`. `DbFile` sounds like a generic database file. This is specific for storing fixed size records though.

src/file.rs
```diff
@@ -0,0 +1,180 @@
...
+	pub file: RwLock<Option<std::fs::File>>,
+	pub path: std::path::PathBuf,
+	pub capacity: AtomicU64,
+	pub dirty: AtomicBool,
+	pub origin: TableId,
+}
+
+impl DbFile {
+	pub fn open(filepath: std::path::PathBuf, header_size: u8, entry_size: u16, origin: TableId) -> Result<Self> {
```
arkpar, line 85: Better name for `header_size` would be `header_entries`, as it is specified in the number of entries. Do we even need it to be a param, or is it always going to be 1?

src/file.rs
```diff
@@ -0,0 +1,180 @@
...
+
+
+const GROW_SIZE: u64 = 256 * 1024;
+
+pub struct DbFile {
+	pub file: RwLock<Option<std::fs::File>>,
+	pub path: std::path::PathBuf,
+	pub capacity: AtomicU64,
+	pub dirty: AtomicBool,
+	pub origin: TableId,
```
arkpar, line 81: Let's just call it `id`

src/file.rs
arkpar: open only preallocates to handle the case when opening an inconsistent database. I.e. there was a power failure and we ended up with a 0-sized file. To replay the log, we need to open the file as empty.
With normal operation `create_file` should be called from within the `grow` function only. Might be a good idea to remove the check for empty file from `write_at`

src/file.rs
```diff
@@ -0,0 +1,180 @@
...
+impl DbFile {
+	pub fn open(filepath: std::path::PathBuf, header_size: u8, entry_size: u16, origin: TableId) -> Result<Self> {
+		let mut capacity = header_size as u64;
+		let file = if std::fs::metadata(&filepath).is_ok() {
+			let file = std::fs::OpenOptions::new().create(true).read(true).write(true).open(filepath.as_path())?;
+			disable_read_ahead(&file)?;
+			if file.metadata()?.len() == 0 {
+				// Preallocate.
+				debug_assert!(header_size as u64 * entry_size as u64 <= GROW_SIZE);
+				file.set_len(GROW_SIZE)?;
```
arkpar, line 93: The file size should not be `GROW_SIZE` exactly, but rather `capacity` * `entry_size`. 
`GROW_SIZE` may contain a non-integer number of records.

arkpar (changes requested):

src/file.rs
cheme: It's used my branch, but actually it is only for the debug assertion, I am removing it.

PR #47, opened 2021-12-03
Testing from db.

cheme:
This PR extracts some changes I was using in https://github.com/cheme/parity-db/tree/fifo_and_set tests.

These allow stopping db in a given state to specifically test the commit overlay, log overlay or db file.
The tests I added are not very useful for the current code base, but here as an example.

In https://github.com/paritytech/parity-db/commit/72b1e1d856731fac0f9f2520f2756f2019f183fb I did try to use it on the stress tests, but only work for small number of commit (I did not debug thoroughly as this was a bit expected from the way it is implemented). Not sure it is really useful to have (it did help me, but can be kept outside of master,  same thing for the pr actually).

### DIFF
--- a/src/db.rs
+++ b/src/db.rs
@@ -150,7 +150,8 @@ impl WaitCondvar<bool> {
 		*work = false;
 	}
 
-	pub fn wait_notify(&self) {
+	#[cfg(test)]
+	fn wait_notify(&self) {
 		let mut work = self.work.lock();
 		self.cv.wait(&mut work)
 	}
--- a/src/db.rs
+++ b/src/db.rs
@@ -127,14 +127,16 @@ struct DbInner {
 }
 
 impl DbInner {
-	fn open(options: &Options, create: bool) -> Result<DbInner> {
+	fn open(options: &Options, create: bool, skip_lock: bool) -> Result<DbInner> {
 		if create {
 			std::fs::create_dir_all(&options.path)?
 		};
 		let mut lock_path: std::path::PathBuf = options.path.clone();
 		lock_path.push("lock");
 		let lock_file = std::fs::OpenOptions::new().create(true).read(true).write(true).open(lock_path.as_path())?;
-		lock_file.try_lock_exclusive().map_err(|e| Error::Locked(e))?;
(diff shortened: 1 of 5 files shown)


arkpar:
I don't think adding test state support all over the code is  worth the extra complexity. We could just add flags that prevent worker threads from starting and then the testing code simply waits for data to appear in the queue or wait on existing condition variables, such as `flush_worker_cv`. This should be enough for unit tests.

cheme:
It is not that far from what it is doing (just here we don't reuse the existing cv but use new ones).
Will keep it as dev tools for now (the tests here are not too useful and when I refactored the btree branch, it may not be that usefull either).

cheme:
I did move this branch to use existing condvar.
Apart from bench code that is very specific and I will probably revert (only for small amount of commits), is this version better? @arkpar

cheme:
The test also does not really make sense, this is really to merge in the btree branch.

arkpar:
> I did move this branch to use existing condvar. Apart from bench code that is very specific and I will probably revert (only for small amount of commits), is this version better? @arkpar

Somewhat. I still think there's no need to expose this to the admin tool. As for `RunMode` I'd call it something like `EnableCommitPipelineStages` and put it inside a struct that's passed to `open_inner`
```rust
struct  InternalOptions {
    create: bool,
    read_only: bool,
    commit_stages: EnableCommitPipelineStages,
    skip_check_lock: bool,
}
```

src/db.rs
```diff
@@ -980,4 +1103,101 @@ mod tests {
...
+			(col_nb, key2.clone(), Some(b"value2".to_vec())),
+			(col_nb, key3.clone(), Some(b"value3".to_vec())),
+		]).unwrap();
+		wait_on.as_ref().map(|w| w.wait_notify());
+		std::mem::drop(db);
+
+		let mut inner_options = InternalOptions::default();
+		inner_options.create = false;
+		inner_options.commit_stages = EnableCommitPipelineStages::CommitOverlay;
+		inner_options.skip_check_lock = true;
```
cheme, line 1188: here

PR #50, opened 2022-02-02
Simplified options a bit

arkpar:
* Removed size tier config as it makes little sense with 256 tiers
* Refactored metadata serialization API.
* Introduced an additional error type when DB is not found.


src/btree/mod.rs
```diff
@@ -174,7 +174,7 @@ impl BTreeTable {
 			id,
 			tables: RwLock::new(values),
 			ref_counted: options.ref_counted,
-			compression: Compress::new(options.compression, options.compression_treshold),
+			compression: Compress::new(options.compression, options.compression_threshold),
```
cheme, line 177: I did see this lately (I am the one that mispell this word a lot), but forgot to fix it. Thanks.

cheme (approved):
LGTM

PR #56, opened 2022-03-03
Change tests to use no threads.

cheme:
Also added fdlimit raise on test_random, and added `process_reindex` on shutdown.

### DIFF
--- a/src/db.rs
+++ b/src/db.rs
@@ -123,7 +123,7 @@ impl WaitCondvar<bool> {
 	fn signal(&self) {
 		let mut work = self.work.lock();
 		*work = true;
-		self.cv.notify_all();
+		self.cv.notify_one();
 	}
 
 	pub fn wait(&self) {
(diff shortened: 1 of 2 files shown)


src/db.rs
```diff
@@ -710,6 +704,7 @@ impl DbInner {
 		while self.enact_logs(false)? {};
 		self.flush_logs(0)?;
 		while self.process_commits()? {};
+		while self.process_reindex()? {};
```
cheme, line 707: @arkpar this strikes me as needed when writing this, but can also be cause for slow shutdown. So not an innocent change.

arkpar: A single call to `process_reindex` does not complete the whole reindex, but just a single batch. I think it is fine to leave reindexing unfinished and restart on next startup, as it may be slow indeed.

arkpar: Ah, did not notice a loop. Anyway, I think finishing the reindex may indeed be too slow. Causing the user to force-kill if no additional warning is given.

PR #57, opened 2022-03-05
Fix a truckload of clippy annotations, warnings and errors.

hirschenberger:
No offense, I know changing codestyle related things may offend devs. Just pick what seems to make sense to you.

Removing superfluous clones and references do make more sense.

But there are other things that seem to be serious, e.g. the unbound locks like this:

`let _ = Self::trigger_reindex(tables, reindex, self.path.as_path());`

### DIFF
--- a/src/db.rs
+++ b/src/db.rs
@@ -144,16 +144,16 @@ impl DbInner {
 		lock_path.push("lock");
 		let lock_file = std::fs::OpenOptions::new().create(true).read(true).write(true).open(lock_path.as_path())?;
 		if !inner_options.skip_check_lock {
-			lock_file.try_lock_exclusive().map_err(|e| Error::Locked(e))?;
+			lock_file.try_lock_exclusive().map_err(Error::Locked)?;
 		}
 
 		let metadata = options.load_and_validate_metadata(inner_options.create)?;
 		let mut columns = Vec::with_capacity(metadata.columns.len());
 		let mut commit_overlay = Vec::with_capacity(metadata.columns.len());
-		let log = Log::open(&options)?;
+		let log = Log::open(options)?;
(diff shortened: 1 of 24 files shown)


src/db.rs
```diff
@@ -1092,13 +1086,9 @@ impl IndexedChangeSet {
 			},
 		};
 		for (key, value) in self.changes.iter() {
-			match column.write_plan(key, value.as_ref().map(|v| v.as_slice()), writer)? {
-				// Reindex has triggered another reindex.
```
arkpar, line 1096: This comment is lost

src/stats.rs
```diff
@@ -88,12 +88,12 @@ impl ColumnStats {
 	pub fn from_slice(data: &[u8]) -> ColumnStats {
 		let mut cursor = Cursor::new(data);
 		let mut value_histogram: [AtomicU32; HISTOGRAM_BUCKETS] = unsafe { MaybeUninit::uninit().assume_init() };
-		for n in 0 .. HISTOGRAM_BUCKETS {
-			value_histogram[n] = read_u32(&mut cursor);
+		for item in value_histogram.iter_mut().take(HISTOGRAM_BUCKETS) {
```
arkpar, line 91: `take` is not need here it seems

arkpar:
Thank you for the PR. Changes look good, apart from a couple of small nits. Formatting uses spaces instead of tabs for some of the new lines, but we are about to adopt cargofmt anyway, so that's not much of an issue.

Regarding `trigger_reindex`, dropping the lock won't cause any issues in both cases. Holding the lock is more robust however.

hirschenberger:
I'll fix the nits and reformat.

PR #58, opened 2022-03-05
More clippy fixes

arkpar:


src/index.rs
```diff
@@ -39,6 +39,9 @@ const EMPTY_CHUNK: Chunk = [0u8; CHUNK_LEN];
 
 pub type Chunk = [u8; CHUNK_LEN];
 
+// TODO: compile time assert
+// (META_SIZE >= HEADER_SIZE + stats::TOTAL_SIZE);
```
cheme, line 43: Had a quick look only found crate 'static_assertions' for it, or use https://github.com/rust-lang/rust/commit/c2dad1c6b9f9636198d7c561b47a2974f5103f6d without the type ascryption.

cheme: Ok did not see there was already #59 :)

PR #59, opened 2022-03-06
Add static_assertions crate and add const assertion

hirschenberger:


arkpar:
Thank you for the PR, but we want to keep dependency list at an absolute minimum. Introducing another dependency for a trivial thing goes agains that. Perhaps a panicking const fn can be used instead of an external crate? Something like
```rust
const META_SIZE: usize = {
	let size = 16 * 1024; // Contains header and column stats
	if HEADER_SIZE + stats::TOTAL_SIZE > size {
		panic!("META_SIZE is too small");
	}
	size
}
```

hirschenberger:
I removed the ext. crate and added a barebones const_assert macro, how about this?

arkpar:
Looks good

PR #65, opened 2022-05-02
Fix a potential deadlock with parallel commits.

arkpar:
If there are multiple threads waiting for the commit queue only one of them would be waken up when the queue is processed.

### DIFF
--- a/admin/src/bench/mod.rs
+++ b/admin/src/bench/mod.rs
@@ -214,7 +214,6 @@ fn writer<D: BenchDb>(
 	let commit_size = COMMIT_SIZE;
 	let mut commit = Vec::with_capacity(commit_size);
 
-
 	loop {
 		let n = COMMITS.fetch_add(1, Ordering::SeqCst);
 		let mut key = n as u64 * COMMIT_SIZE as u64 + offset;
@@ -234,7 +233,7 @@ fn writer<D: BenchDb>(
 		commit.clear();
 
 		if n >= start_commit + args.commits || shutdown.load(Ordering::Relaxed) {
-			break;
+			break
 		}
 	}
 }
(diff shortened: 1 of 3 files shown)


admin/src/bench/mod.rs
```diff
@@ -210,14 +211,12 @@ fn writer<D: BenchDb>(
...
 	let commit_size = COMMIT_SIZE;
 	let mut commit = Vec::with_capacity(commit_size);
 
-	for n in start_commit..start_commit + args.commits {
-		if shutdown.load(Ordering::Relaxed) {
-			break
-		}
+	loop {
+		let n = COMMITS.fetch_add(1, Ordering::SeqCst);
+		let mut key = n as u64 * COMMIT_SIZE as u64 + offset;
```
cheme, line 219:
```suggestion
		let mut key = if n == start_commit {
			n as u64 * COMMIT_SIZE as u64 + offset
		} else {
			n as u64 * COMMIT_SIZE as u64
		};
```

admin/src/bench/mod.rs
```diff
@@ -231,16 +230,38 @@ fn writer<D: BenchDb>(
...
+	seed: u64,
+	index: u64,
+	shutdown: Arc<AtomicBool>,
+) {
+	// Query random keys while writing
+	let mut rng = rand::rngs::SmallRng::seed_from_u64(seed + index);
 	while !shutdown.load(Ordering::Relaxed) {
-		thread::sleep(std::time::Duration::from_millis(500));
+		let commits = COMMITS.load(Ordering::Relaxed) as u64;
+		if commits == 0 {
```
cheme, line 252: start_commit rather than 0? Actually since start_commit is restart index, there should be content so probably good like this.

cheme (approved):
So idea is if multiple commit are stuck on full queue, and only one get released, so change sounds good to me.

Thinking out loud, these kind of concurrent write access are released in random order, so it implies they are call in random order. If they were not call in random order, there will be no concurrent accesses in the first place so it is fine to release this way.

Also previously code was using notify_all, I am the one that introduce this issue when reverting some code badly. So definitely safe to have.

admin/src/bench/mod.rs
arkpar: why?

admin/src/bench/mod.rs
arkpar: `COMMITS` is the number of commits in the database. It should be possible to query any of them.

cheme: yes thought so, certainly no use to target specifically the latest ones.

PR #68, opened 2022-05-06
Fix locking order in BtreeIterator::next

arkpar:


cheme (approved):
Also observe fix on bench.

PR #69, opened 2022-05-06
Does not fix anything: just removes two benign things that raise unnecessary questions.

RGafiyatullin:
* unnecessarily public function;
* unneeded explicit scope: probably an echo of some previous refactoring.


cheme (approved):
thanks

PR #70, opened 2022-05-11
Fix potential deadlock on double reindex

arkpar:
Also made commit counting in the bench tool more precise.

### DIFF
--- a/admin/src/bench/mod.rs
+++ b/admin/src/bench/mod.rs
@@ -32,6 +32,7 @@ use std::{
 };
 
 static COMMITS: AtomicUsize = AtomicUsize::new(0);
+static NEXT_COMMIT: AtomicUsize = AtomicUsize::new(0);
 static QUERIES_HIT: AtomicUsize = AtomicUsize::new(0);
 static QUERIES_MISS: AtomicUsize = AtomicUsize::new(0);
 static ITERATIONS: AtomicUsize = AtomicUsize::new(0);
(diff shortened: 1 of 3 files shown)


admin/src/bench/mod.rs
```diff
@@ -192,7 +193,11 @@ fn writer(
 	let mut commit = Vec::with_capacity(commit_size);
 
 	loop {
-		let n = COMMITS.fetch_add(1, Ordering::SeqCst);
+		let n = NEXT_COMMIT.fetch_add(1, Ordering::SeqCst);
```
cheme, line 196: Do we need NEXT_COMMIT here or just COMMITS.load + 1 could work?

src/column.rs
```diff
@@ -397,14 +397,25 @@ impl HashColumn {
...
 			return Ok(PlanOutcome::Skipped)
 		}
 		match tables.index.write_insert_plan(key, address, None, log)? {
 			PlanOutcome::NeedReindex => {
-				log::debug!(target: "parity-db", "{}: Index chunk full {}", tables.index.id, hex(key));
-				let _lock = Self::trigger_reindex(tables, reindex, self.path.as_path());
-				self.write_reindex_plan(key, address, log)?;
+				log::debug!(target: "parity-db", "{}: Index chunk full {} when reindexing", tables.index.id, hex(key));
+				let (tables, reindex) = Self::trigger_reindex(tables, reindex, self.path.as_path());
+				self.write_reindex_plan_locked(tables, reindex, key, address, log)?;
```
cheme, line 418: make sense

cheme (approved):

admin/src/bench/mod.rs
cheme: ah, ok , forgot the fact there is multiple writers, it is indeed needed

PR #84, opened 2022-07-21
Added clear_column funtion

arkpar:


cheme (approved):
LGTM, maybe something like https://github.com/paritytech/parity-db/commit/ad1a9c0191ba168846388e5bb1a0d9ff746a621f could be added (so we fail in these cases).

PR #87, opened 2022-07-28
Expose basic stats

### LINKED ISSUE #80: Expose column stats
Current stats can be written in text form, but it would be useful to access them programmatically too. For instance, I'm interested in number of elements stored in a hash column and/or amount of data.

The data is already there, but no API to access it unfortunately.

arkpar:
Closes #80


cheme (approved):
Changes looks good to me.
Not entirely sure why we only expose a summary of stats (I am assuming it makes the api more stable and agree some infos may not be relevant).

arkpar:
I imagine most of the stats are of low iterest to API uses. We can expand it later if required

PR #88, opened 2022-08-12
Increment Rc and Change enum

cheme:
I did get some code from old branch to propose this PR.

It expose a new operand to commit: `IncRc` that allow increase rc count in columns using rc, but without knowing the value.

This is not very important feature (can be use at one place in chain storage of substrate).

The PR also define a named enum `Change` that could be extended with different operand (I did have this in multiple branch so it make sense for me to have it).

### DIFF
--- a/src/column.rs
+++ b/src/column.rs
@@ -17,7 +17,7 @@
 use crate::{
 	btree::BTreeTable,
 	compress::Compress,
-	db::check::CheckDisplay,
+	db::{check::CheckDisplay, Change},
 	display::hex,
 	error::{Error, Result},
 	index::{Address, IndexTable, PlanOutcome, TableId as IndexTableId},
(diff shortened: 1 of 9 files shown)


src/column.rs
```diff
@@ -469,65 +469,70 @@ impl HashColumn {
...
+					Ok(r)
+				},
+				Change::RemoveValue(key) => {
+					log::trace!(target: "parity-db", "{}: Deleting missing key {}", tables.index.id, hex(key));
+					if self.collect_stats {
+						self.stats.remove_miss();
+					}
+					Ok(PlanOutcome::Skipped)
+				},
+				Change::IncRc(key) => {
```
arkpar, line 493: This should be logged in a new stats field. 
There's an important difference between using `SetValue` and `IncRc`. The code might check that the value exists and try to increase the reference with `IncRc`, while some other thread commits a transaction that removes the value. In that case `IncRc` will fail, since the value does not exist anymore, whils `SetValue` would work fine. The user of the `IncRc` have to be 100% sure they don't get the value deleted before the trasaction is commited. Having a stat field might help diagnose such issues.

src/db.rs
```diff
@@ -1100,6 +1121,56 @@ impl CommitOverlay {
 	}
 }
 
+#[derive(PartialEq, Eq)]
+pub enum Change<Key, Value> {
```
arkpar, line 1125: Since this is user facing, I'd call it `Operation`

src/db.rs
```diff
@@ -1100,6 +1121,56 @@ impl CommitOverlay {
...
+	SetValue(Key, Value),
```
arkpar, line 1126: Simply`Set` would work

src/db.rs
```diff
@@ -1100,6 +1121,56 @@ impl CommitOverlay {
...
+	RemoveValue(Key),
```
arkpar, line 1127: `Dereference`

src/db.rs
```diff
@@ -1100,6 +1121,56 @@ impl CommitOverlay {
...
+	IncRc(Key),
```
arkpar, line 1128: `Reference`

src/db.rs
```diff
@@ -1100,6 +1121,56 @@ impl CommitOverlay {
...
+		}
+	}
+
+	pub fn into_key(self) -> Key {
+		match self {
+			Change::SetValue(k, _) | Change::RemoveValue(k) | Change::IncRc(k) => k,
+		}
+	}
+
+	pub fn is_rc_ops(&self) -> bool {
```
arkpar, line 1156: `is_reference_op`

src/options.rs
```diff
@@ -57,6 +57,9 @@ pub struct ColumnOptions {
 	/// Allows for skipping additional key hashing.
 	pub uniform: bool,
 	/// Use reference counting for values.
+	///
+	/// Reference counting do not enforce immediate removal
+	/// and user should not check for missing value.
```
arkpar, line 62: What do you mean here? As soon as reference count reaches 0 the value is removed, isn't it?

src/db.rs
cheme: Here I previously got two different operattion: Remove and DecRc, since they are exactly the same I remove one.
But If I want `Dereference` I would probably need to also use `Remove`. (or we go back to a semantic where Set uses Option)

src/options.rs
cheme: Not in the commit overlay, there we simply skip all removal.

cheme: (I think our rc usecase is related to rc + uniform, which make sense for trie nodes).
Thinking about it I almost thought of forbidding rc without uniform or have two rc behavior (but I don't think there is a use case yet with true rc).

arkpar: Ok, makes sense. We should probably rework commit overlay, so that it also includes RC changes. In `get` If for some key the RC change delta in he overlay is negative, we query the backend for the actual counter value.

src/db.rs
arkpar: One operation is OK, but I think it should be called `Dereference`. For non-ref counted columns you can still think it removes a single reference and removes the value.

src/column.rs
cheme: I've seen you remove the statistic from the flush of the memmap (makes sense).
I was wondering if maybe we would just want the stats in their own file without using the WAL (having to add a single field is not such an happy perspective to me), so we can make it more flexible.
Could for example just use something like crates binary-layout over a single memmap (no idea if the crate is good I just remember I wanted to test it).

arkpar: Keeping it with the index seemed more simple. For now you can just add the field to the end of the struct and it should not break compatibility with existing data.

arkpar (approved):

PR #91, opened 2022-08-29
Add debug implementations to almost all types

i1i1:


arkpar:
Why? Most of these types are crate internal. Since there's no actual logging code added, all this does is slows down compilation.

i1i1:
> Why?

I just found it confusing that I couldn't derive `Debug` for structs which contain parity-db.

> Most of these types are crate internal. Since there's no actual logging code added, all this does is slows down compilation.

Some types are public, though (or contained in some nested way in `parity_db::Db`). For internal types, it would be easier to debug code in future (I personally just added `derive(Debug)` by hands for debugging anyway).

PR #92, opened 2022-09-02
update lz4 to 1.24.0

nbaztec:
The current `lz4:1.23.2` depends upon `lz4-sys:1.9.2`, which is flagged via dependabot [vulnerability](https://github.com/advisories/GHSA-9q5j-jm53-v7vr) alert. As a solution `lz4-sys` must be update to `1.9.4` which is included in `lz4:1.24.0`


cheme (approved):
Thanks for spotting it.

PR #94, opened 2022-09-09
Allow opening db multiple times in read only mode

sandreim:
We need this capability for monitoring db usage for paritydb enable nodes.

https://github.com/paritytech/polkadot-introspector/issues/107

### DIFF
--- a/src/db.rs
+++ b/src/db.rs
@@ -832,7 +832,8 @@ impl Db {
 	}
 
 	pub fn open_read_only(options: &Options) -> Result<Db> {
-		let inner_options = InternalOptions { read_only: true, ..Default::default() };
+		let inner_options =
+			InternalOptions { read_only: true, skip_check_lock: true, ..Default::default() };
 		Self::open_inner(options, &inner_options)
 	}
 
@@ -1476,6 +1477,18 @@ mod tests {
 		let options = Options::with_columns(tmp.path(), 5);
 		assert!(Db::open_or_create(&options).is_ok(), "New database should be created");
 		assert!(Db::open(&options).is_ok(), "Existing database should be reopened");
+	}
+
+	#[test]
+	fn test_db_open_multiple_times_read_only() {
+		let tmp = tempdir().unwrap();
+		let options = Options::with_columns(tmp.path(), 5);
+		let _db = Db::open_or_create(&options).expect("Failed to open db");
+		let _db_ro_1 = Db::open_read_only(&options).expect("Failed to open db in read only mode");
+		let _db_ro_2 = Db::open_read_only(&options).expect("Failed to open db in read only mode");
+
+		// Should fail if we try to open again using write capabilities.
+		assert!(Db::open_or_create(&options).is_err());
 	}
 
 	#[test]


cheme:
I started thinking on how to implement it (wanted to try it this week but got delayed on other stuff).
Issue with having a node that read open file is that the file(s) may be in an incoherent state: some part written and not flushed, some over file written and not flush... So I fear just doing this will lead to the reading process potentially reading incoherent state.
To attach a reading access, I fear we need to share at least the log overlay cache (read node will be a bit behind but in a coherent state). This way we ensure the  state is in right state with a single commit.
But this means at least an ipc to send the commit numbers and do lock during update and log overlay cache memory shared. Or just ipc all items for the log overlay and duplicate the log overlay cache.

Maybe I overthink it and it can run this way (would need testing). CC\ @arkpar

cheme:
@sandreim what are the operation the introspector needs to be able to do ? just peeking at https://github.com/paritytech/polkadot-introspector/blob/master/src/kvdb/paritydb.rs it gives me the feeling it expect iterator available (which is rarely the case for paritydb).

sandreim:
Our usecase is just iterating and measuring key and value sizes. Doesn’t need to be 100% accurate wrt contents.

cheme:
If implementing the read only access on a node open for writing, he introspector will only work with the following columns https://github.com/paritytech/polkadot/blob/5de9b7f6e5608c4e7ac91b9f241064e4fe48cb85/node/service/src/parachains_db/mod.rs#L40 .
For other column the introspector would need us to store things differently:
- attach key with values
- do full scan iteration of the column (even for a single prefix)
- iterate offline (we currently only iterate on non index column in migration process), or do the fullscan iteration over the node overlay to. This point is not a real blocker.

On a side note I remember seeing the need for read access in the context of substrate archive, so I think it could make sense to have it.

vstakhov:
Another option is to have an iterator but return a hash instead of the key. It would be enough for our use cases (we just need db size in the majority of the cases). If we can get a size of DB per column somehow without iterating, I'm open for adding such an optimization here.

arkpar:
Would simply adding the size of DB files on disk work? It is much faster than iterating and won't hinder the system IO.

cheme:
We also can probably got column stats (I would read stats directly from the files (no db opening) in this case) when they are active.

sandreim:
> We also can probably got column stats (I would read stats directly from the files (no db opening) in this case) when they are active.

I think this would do the job.

sandreim:
> Would simply adding the size of DB files on disk work? It is much faster than iterating and won't hinder the system IO.

Can we compute the column sizes based on the file sizes? If so, then yeah this is a good alternative as well, but more error prone if on disk structure changes

vstakhov:
> Would simply adding the size of DB files on disk work? It is much faster than iterating and won't hinder the system IO.

For now, we have 3 parameters per column:

* Keys count
* Keys size
* Values size

If we can have that from metadata, that would be great for these purposes. And it will be much faster than RocksDB option which is also good.

arkpar:
We can add a function to estimate DB size based on the size of the index and value tables. This will give you total data size, but not the number of key-values. To get more detailed info, we can add a function to get stats without opening the database. However there's a couple of catches.
1. You would need to enable stat collection when opening the database for in your main process. This comes with small performance hit.
2. Stats are currently flushed only when closing the database.

Allowing opening a database for shared access as this PR currently tries to do is indeed pronce to misuse. As @cheme noticed we can't guarantee data consistency when another instance is writing.

arkpar:
I'm closing this fo reasons listed in he [comment](https://github.com/paritytech/parity-db/pull/94#issuecomment-1245193637)

PR #96, opened 2022-09-13
Upgrades jemallocator and uses it anywhere but msvc

Tpt:
The targets in the code vs the Cargo.toml were different


arkpar (approved):

cheme (approved):
lgtm (I guess musl build did not work before).

PR #97, opened 2022-09-13
Moves IdentityHash out of lib.rs

Tpt:
Simplifies the lib.rs file

Inlines the const_assert macro that is only used once

### DIFF
--- a/src/hash.rs
+++ b/src/hash.rs
@@ -1,0 +1,18 @@
+// Copyright 2021-2022 Parity Technologies (UK) Ltd.
+// This file is dual-licensed as Apache-2.0 or MIT.
+
+use std::hash::BuildHasherDefault;
+
+#[derive(Default)]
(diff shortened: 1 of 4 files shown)


arkpar (approved):

src/hash.rs
```diff
@@ -0,0 +1,18 @@
...
+
+#[derive(Default)]
+pub struct IdentityKeyHash(u64);
+pub type IdentityBuildHasher = BuildHasherDefault<IdentityKeyHash>;
+
+impl std::hash::Hasher for IdentityKeyHash {
+	fn finish(&self) -> u64 {
+		self.0
+	}
+
```
cheme, line 14: you did not like my unreachables :) ? I guess it was a safe guard to avoid someone using the hasher incorrectly, but it takes loc, I am fine with removing them (will panic anyway).

cheme (approved):

PR #98, opened 2022-09-13
Runs address sanitizer as part of the CI

Tpt:
Instrumentation provided by LLVM to track bad memory access

Useful when unsafe code is used

Allows to also run tests on nightly

Documentation: https://doc.rust-lang.org/beta/unstable-book/compiler-flags/sanitizer.html#addresssanitizer


arkpar (approved):

cheme (approved):
Nice to have :+1:

PR #99, opened 2022-09-13
Adds compression threshold to column options serialization

Tpt:


arkpar:
It is omitted from serialization because it does not really affect how existing data is handled. Changing it does not break compatibility. I'd argue it should not even be in the metadata struct, but just in `ColumnOptions`. @cheme what do you think?

cheme:
> It is omitted from serialization because it does not really affect how existing data is handled. Changing it does not break compatibility. I'd argue it should not even be in the metadata struct, but just in `ColumnOptions`. @cheme what do you think?

Yes, I agree, should be modifiable when we instantiate the db.

Tpt:
Thank you for the review. My motivation for this change was to allow opening databases with non-default compression threshold. 

>  I'd argue it should not even be in the metadata struct, but just in ColumnOptions.

Ok! So we might need to have a `ColumnMetadata` struct that woud keep the fixed metadata (i.e. everything in `ColumnOptions` but `compression_threshold`). Am I correct?

Tpt:
> I suggest simply moving `compression_threshold` from `ColumnOptions` to `Options` as a `HashMap<ColId, u32>`.

Done in #103

PR #102, opened 2022-09-13
Remove incorrect assertion

### LINKED ISSUE #101: Panic when doing duplicated references in the same transaction on a ref-counted btree
The following test:
```rust
#[test]
fn test_duplicated_reference_on_ref_counted_btree() {
	let tmp = tempdir().unwrap();
	options.columns[0].ref_counted = true;
	db.commit_changes(vec![(0, Operation::Reference(vec![255])), (0, Operation::Reference(vec![255]))]).unwrap();
}
```

Fails with a panic:

```
thread 'db::tests::test_duplicated_reference_' panicked at 'assertion failed: changes[0].key() < changes[1].key()', src/btree/node.rs:76:17
```

(first test found with fuzzing!)

cheme:
Removed assertion was incorrect, we skip Set operation on same key so we should not have two consecutive key in this case.
On the other hand whith reference operation on does not overwrite the other, so it make sense to have two consecutive identic keys (can also be key from a set operation).
I also did skip some Set operation when reference is in between to set operation but that's a corner case.

Makes me think reference operation could be optimized to only increase reference once (can be a bit tricky as if at any point the Rc is bellow 0, then folowing calls need to be skipped unless value written again, core pb is that we don't know the rc value when we want to skip such operation and probably don't want to fetch the rc, so should be optimizable but at a different location).

This issue should not happen in hash indexed tables: the change set is not sorted and we don't skip operations.

fix #101

### DIFF
--- a/src/btree/node.rs
+++ b/src/btree/node.rs
@@ -64,11 +64,19 @@ impl Node {
 		log: &mut LogWriter,
 	) -> Result<(Option<(Separator, Child)>, bool)> {
 		loop {
-			if changes.len() > 1 {
-				if changes[0].key() == changes[1].key() &&
-					!changes[0].is_reference_ops() &&
-					!changes[1].is_reference_ops()
-				{
+			if changes.len() > 1 && !changes[0].is_reference_ops() {
+				let mut skip = false;
+				for next in changes[1..].iter() {
+					if changes[0].key() == next.key() {
+						if !next.is_reference_ops() {
+							skip = true;
+							break
+						}
+					} else {
+						break
+					}
+				}
+				if skip {
 					// TODO only when advancing (here rec call useless)
 					*changes = &changes[1..];
 					continue
--- a/src/btree/node.rs
+++ b/src/btree/node.rs
@@ -73,7 +73,6 @@ impl Node {
 					*changes = &changes[1..];
 					continue
 				}
-				debug_assert!(changes[0].key() < changes[1].key());
 			}
 			let r = match &changes[0] {
 				Operation::Set(key, value) =>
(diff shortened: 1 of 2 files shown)


src/btree/node.rs
```diff
@@ -64,16 +64,23 @@ impl Node {
...
+			if changes.len() > 1 && !changes[0].is_reference_ops() {
+				let mut skip = false;
+				for next in changes[1..].iter() {
+					if changes[0].key() == next.key() {
+						if !next.is_reference_ops() {
+							skip = true;
+							break
+						}
+					} else {
+						break
```
Tpt, line 76: Shouldn't we `continue` here instead of `break` because it is fine to skip the operation if there is an operation that overrides the current one it is fine if it is not consecutive? Or is there some bad interactions with reference ops?

Tpt (approved):

src/btree/node.rs
cheme: I love github, I am not sure if line 76 or 73, for 76 it means we are on a different key so we check no other Set operation coming up (keys are sorted), for 73 we did see a Set operation afterward so we can skip this current set operation and we don't need to look further.
I mean the idea is for changes at the same key, all Set before the last one can be skip. Reference (does rc + 1) or dereference (rc - 1) cannot as we need the sum of these calls.
(and doing the sum at this level is not doable because we don't want to fetch the current rc and after rc reach 0 any following reference operation are noops (unless there was a Set in between)). Actually we could safely merge consecutive Reference and even Dereference as long as there is no chance to reach 0, but would need to change some Reference enum to contains a counter (which would not be bad).

Tpt: > If break at line 76 it means we are on a different key so we check no other Set operation coming up (keys are sorted),

Ok! Thank you! I missed that the key were sorted.

arkpar (approved):

PR #106, opened 2022-09-14
Adds basic model-based fuzzers

Tpt:
Checks that the database behaves like an in-memory collection when a sequence of transactions and restarts are applied.

Covered cases:
- hash map and btree (for btree iteration order is checked)
- the 3 compressions
- reference counting or not

### DIFF
--- a/fuzz/fuzz_targets/refcounted_model.rs
+++ b/fuzz/fuzz_targets/refcounted_model.rs
@@ -1,0 +1,131 @@
+// Copyright 2021-2022 Parity Technologies (UK) Ltd.
+// This file is dual-licensed as Apache-2.0 or MIT.
+
+//! Model check with reference counting:
+//! Checks that a sequence of operations and restarts behaves the same as an in-memory collection.
+
+#![no_main]
+use arbitrary::Arbitrary;
+use libfuzzer_sys::fuzz_target;
+use std::collections::btree_map::{BTreeMap, Entry};
+use tempfile::tempdir;
+
+#[derive(Arbitrary, Debug, Clone, Copy)]
+enum Operation {
+	Set(u8, u8),
+	Dereference(u8),
+	Reference(u8),
+}
+
+#[derive(Arbitrary, Debug)]
+enum Action {
+	Transaction(Vec<Operation>),
+	Restart,
+}
+
+#[derive(Arbitrary, Debug, Clone, Copy)]
+enum CompressionType {
+	NoCompression,
+	Snappy,
+	Lz4,
+}
+
+#[derive(Arbitrary, Debug)]
+struct Config {
+	btree_index: bool,
+	compression: CompressionType,
+}
+
+fuzz_target!(|entry: (Config, Vec<Action>)| {
+	let (config, actions) = entry;
+	let dir = tempdir().unwrap();
+	let options = parity_db::Options {
+		path: dir.path().to_owned(),
+		columns: vec![parity_db::ColumnOptions {
+			ref_counted: true,
+			compression: match config.compression {
+				CompressionType::NoCompression => parity_db::CompressionType::NoCompression,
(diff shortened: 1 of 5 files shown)


arkpar (approved):

fuzz/fuzz_targets/simple_model.rs
```diff
@@ -0,0 +1,102 @@
...
+#![no_main]
+use arbitrary::Arbitrary;
+use libfuzzer_sys::fuzz_target;
+use std::collections::btree_map::BTreeMap;
+use tempfile::tempdir;
+
+#[derive(Arbitrary, Debug)]
+enum Action {
+	Transaction(Vec<(u8, Option<u8>)>),
+	Restart,
```
cheme, line 16: I never did use Arbitrary, looks nice.
I could probably rewrite some code with it.

fuzz/fuzz_targets/refcounted_model.rs
```diff
@@ -0,0 +1,131 @@
...
+#[derive(Arbitrary, Debug)]
+enum Action {
+	Transaction(Vec<Operation>),
+	Restart,
+}
+
+#[derive(Arbitrary, Debug, Clone, Copy)]
+enum CompressionType {
+	NoCompression,
+	Snappy,
```
cheme, line 29: We could also have only one compressed variant (Lz4).

fuzz/fuzz_targets/refcounted_model.rs
```diff
@@ -0,0 +1,131 @@
...
+	Lz4,
+}
+
+#[derive(Arbitrary, Debug)]
+struct Config {
```
cheme, line 34: We could also use different fuzz_target for different config, but if config get complex it won't be manageable so sounds ok.
(if needed it would not take long to use a static config)

fuzz/fuzz_targets/simple_model.rs
```diff
@@ -0,0 +1,102 @@
...
+			assert_eq!(db_content, model_content);
+
+			// We check the BTree backward iterator
+			model_content.reverse();
+			db_iter.seek_to_last().unwrap();
+			let mut db_content = Vec::new();
+			while let Some((k, v)) = db_iter.prev().unwrap() {
+				db_content.push((k[0], v[0]));
+			}
+			assert_eq!(db_content, model_content);
```
cheme, line 99: possible checks for iterator is changing the direction of the iterator before reaching last.
But I am not too sure iterator behavior on change direction should remain this way (got a branch that change it but I need to apply a recursive stack overflow fix on it first).

cheme (approved):

cheme (approved):

cheme (approved):

cheme (approved):

cheme (approved):

fuzz/fuzz_targets/simple_model.rs
Tpt: Thank you! What about keeping it this way for now and add deeper tests later if still relevant?

fuzz/fuzz_targets/refcounted_model.rs
Tpt: Yes, might be nice. But I believe we will quickly want to play with various thresholds (queue size...) to do deeper testing and so a dynamic configuration makes more sense. What do you think about it?

fuzz/fuzz_targets/refcounted_model.rs
Tpt: Indeed. But I'm not sure it harms to test both variants.

fuzz/fuzz_targets/simple_model.rs
cheme: yes sure.

PR #108, opened 2022-09-14
allow using ref count for btree column

### LINKED ISSUE #104: Values do not seems to be persisted on ref-counted BTree
Here is a failing test I believe should pass:

```rust
	#[test]
	fn fuzz_test() {
		let dir = tempdir().unwrap();
		let options = Options {
			path: dir.path().to_owned(),
			columns: vec![ColumnOptions {
				ref_counted: true,
				btree_index: true,
				..ColumnOptions::default()
			}],
			sync_wal: true,
			sync_data: true,
			stats: false,
			salt: None,
		};

		db.commit_changes(vec![(0, Operation::Set(vec![0], vec![0]))]).unwrap();

		// The value is there before closing and opening again
		assert_eq!(db.get(0, &[0]).unwrap(), Some(vec![0]));

		// Ensures we should properly join on shutdown
		assert!(db.join_on_shutdown);

		// The value is not there anymore
		drop(db);
		let db = Db::open_read_only(&options).unwrap();
		assert_eq!(db.get(0, &[0]).unwrap(), Some(vec![0]));
	}
```

cheme:
Fix #104 

BTree columns were never used before with ref counted. My first idea was to forbid it, but it was fairly implementable I think.

A new parameter was added to write_existing_value to be able to use rc logic only on value (and not on indexing nodes).
While debugging I also found that root was uslessly written twice, so I did remove a write_plan function (second time was noop since changes where already writte).
Also had to disable node removal when the rc change did not delete value.
Also removed the node parent of the value marked as modified if the value index did not change.

### DIFF
--- a/src/btree/node.rs
+++ b/src/btree/node.rs
@@ -64,23 +64,13 @@ impl Node {
 		log: &mut LogWriter,
 	) -> Result<(Option<(Separator, Child)>, bool)> {
 		loop {
-			if changes.len() > 1 && !changes[0].is_reference_ops() {
-				let mut skip = false;
-				for next in changes[1..].iter() {
-					if changes[0].key() == next.key() {
-						if !next.is_reference_ops() {
-							skip = true;
-							break
-						}
-					} else {
-						break
-					}
-				}
-				if skip {
-					// TODO only when advancing (here rec call useless)
-					*changes = &changes[1..];
-					continue
-				}
+			if !btree.ref_counted &&
+				changes.len() > 1 &&
+				!matches!(changes[1], Operation::Reference(..)) &&
+				changes[0].key() == changes[1].key()
+			{
+				*changes = &changes[1..];
+				continue
 			}
 			let r = match &changes[0] {
 				Operation::Set(key, value) =>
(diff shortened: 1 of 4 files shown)


src/btree/node.rs
```diff
@@ -260,30 +250,35 @@ impl Node {
...
-					self.write_child(i, child, values, log)?;
-					if let Some(sep) = sep {
-						self.set_separator(i, sep);
-					}
-					if need_balance {
-						self.rebalance(depth, i, values, log)?;
+					false,
+				)?
+				.0
+				.is_none()
```
cheme, line 263: return type of this function is pretty terrible (multiple meaning depending on the way the function is called), yet I did not change it in this pr.

src/column.rs
```diff
@@ -886,6 +887,7 @@ impl Column {
 		change: &Operation<K, V>,
 		log: &mut LogWriter,
 		stats: Option<&ColumnStats>,
+		btree_index_node: bool,
```
arkpar, line 890: Maybe just call this `ref_counted` and simply check `if ref_counted` instead of `if tables.ref_counted && !btree_index_node`?

cheme: :facepalm: clearly yes.

arkpar (approved):

src/db.rs
```diff
@@ -1854,12 +1858,47 @@ mod tests {
...
+						},
+				}
+				if insert {
+					state.insert(k.clone(), v.clone().map(|v| (v, 1)));
+				}
+				if remove {
+					state.remove(k);
+				}
+			}
+			for (key, value) in state.into_iter() {
```
Tpt, line 1887: Nitpicking: `.into_iter()` is not needed here. It's what for loop does by default.

src/db.rs
```diff
@@ -1854,12 +1858,47 @@ mod tests {
 				expected_count -= 1;
 			}
 		}
-		assert_eq!(col_stats.total_values, expected_count);
+		if ref_counted {
+			let mut state: BTreeMap<Vec<u8>, Option<(Vec<u8>, usize)>> = Default::default();
```
Tpt, line 1862: Genuine question: What is the difference in `states` between a `None` value and the value not being in the `BTreeMap`? I would believe that `BTreeMap<Vec<u8>, (Vec<u8>, usize)>` would be as good but I'm maybe missing something.

Tpt (approved):
I don't feel knowledgeable enough yet of the BTree internal implementation to do a proper review. I just have some very minor nitpicks about the test code. Feel free to ignore them.

src/db.rs
cheme: I ask me the same thing yesterday (actually did start changing it). Having None is needed to assert the value a few line bellow do not exists in the db (so if a removal fail we would see the value still in db).

src/db.rs
cheme: Thanks, old habit I want to change, funny that's something I would expect clippy to find.

PR #109, opened 2022-09-14
Last state iterator

cheme:
This is a small refactor of iterator logic when switching backward/forward.
Mainly removea bit of redundant code and makes that we do not return same value when changing iteration direction.

Currently draft as I need to reread it and clippy it, but @Tpt it may be interesting to run your fuzzer on this branch.

### DIFF
--- a/src/btree/iter.rs
+++ b/src/btree/iter.rs
@@ -14,9 +14,18 @@ use crate::{
 use parking_lot::RwLock;
 
 #[derive(Debug, PartialEq, Eq, PartialOrd, Ord, Clone, Copy)]
-pub enum SeekTo {
-	Include,
-	Exclude,
+pub enum SeekTo<'a> {
+	Include(&'a [u8]),
+	Exclude(&'a [u8]),
+}
+
+impl<'a> SeekTo<'a> {
+	pub fn key(&self) -> &'a [u8] {
+		match self {
+			SeekTo::Include(key) => key,
+			SeekTo::Exclude(key) => key,
+		}
+	}
 }
 
 #[derive(Debug)]
@@ -29,6 +38,8 @@ pub struct BTreeIterator<'a> {
 	pending_backend: Option<PendingBackend>,
 	last_key: LastKey,
 }
+
+type IterResult = Result<Option<(Vec<u8>, Vec<u8>)>>;
 
 #[derive(Debug)]
 struct PendingBackend {
@@ -86,11 +97,10 @@ impl<'a> BTreeIterator<'a> {
 		self.last_key = LastKey::Seeked(key.to_vec());
 		self.pending_backend = None;
 		self.seek_backend(
-			key,
+			SeekTo::Include(key),
 			record_id,
 			self.table,
 			&*log,
-			SeekTo::Include,
 			IterDirection::Forward,
 		)
 	}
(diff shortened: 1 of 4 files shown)


cheme:
Nice, to hear. Currently cleaning this branch.

cheme:
This PR should be reviewable.

@i1i1 with this pr, I did change the way the iterator behave when switching direction.
on a set of key a, b, c. and with calls next, next, prev, perf
before we had a -> b -> b -> a
now we have a -> b -> a -> nil
it ends up being easier to fuzz (same behavior as btreemap).

src/btree/iter.rs
```diff
@@ -344,79 +353,121 @@ impl BTreeIterState {
...
-					Some(child) => {
-						let last_separator = child.last_separator_index().unwrap_or_default();
-						let last_child = child.last_child_index().unwrap_or_default();
-						if *ix > 0 {
-							*ix -= 1;
+		loop {
+			let is_leaf = btree.depth as usize + 1 == self.state.len();
+			if let Some(state) = self.state.last_mut() {
+				let next = match (direction, &state.0) {
+					(_, LastIndex::Descend(_)) => unreachable!("exit function clean it"),
```
arkpar, line 370:
```suggestion
					(_, LastIndex::Descend(_)) => unreachable!("Las state should be clean after calling self.exit"),
```

Is it always cleaned by `self.exit` though? From what I can follow, sometimes `self.enter` is called instead.

src/db.rs
```diff
@@ -1083,48 +1085,51 @@ impl CommitOverlay {
...
-			return self.btree_indexed.iter().next().map(|(k, (_, v))| (k.clone(), v.clone()))
-		};
-
-		let mut iter = self.btree_indexed.range::<Vec<u8>, _>(key..);
-
-		let (k, v) = if let Some((k, (_, v))) = iter.next() { (k, v) } else { return None };
-
-		if from_seek || k != key {
-			return Some((k.clone(), v.clone()))
+	pub fn btree_next(&self, last_key: &crate::btree::LastKey) -> Option<(Value, Option<Value>)> {
```
Tpt, line 1088: I believe might be possible to return slices instead of `Value` here but it's likely to not be very useful anyway...

src/btree/iter.rs
```diff
@@ -68,232 +94,178 @@ impl<'a> BTreeIterator<'a> {
...
+					(IterDirection::Backward, std::cmp::Ordering::Greater) |
+					(IterDirection::Forward, std::cmp::Ordering::Less) => {
+						self.pending_backend = Some(PendingBackend {
+							next_item: Some((backend_key, backend_value)),
+							direction,
+						});
+						if let Some(value) = commit_value {
+							Some((commit_key, value))
+						} else {
+							std::mem::drop(log);
```
Tpt, line 176: log is going to be dropped anyway just afterward (same for the other `return Err(())` places)

src/btree/iter.rs
```diff
@@ -68,232 +94,178 @@ impl<'a> BTreeIterator<'a> {
...
+	pub fn prev(&mut self) -> IterResult {
+		loop {
+			if let Ok(result) = self.iter_inner(IterDirection::Backward) {
+				return result
 			}
 		}
 	}
 
-	pub fn prev(&mut self) -> Result<Option<(Vec<u8>, Vec<u8>)>> {
+	fn iter_inner(&mut self, direction: IterDirection) -> std::result::Result<IterResult, ()> {
```
Tpt, line 136: I believe it would make the code more clear to move the `loop` from `prev` and `next` here and have explicit `continue` call instead of returning `Err`. It took me some time to figure out that returning `Err` is not a falure but a retry... But it's code style so feel free to ignore it...

arkpar (approved):

src/btree/iter.rs
```diff
@@ -321,20 +293,57 @@ impl<'a> BTreeIterator<'a> {
...
+					self.state.clear(); // should actually be unreachable
+					self.fetch_root = false;
+					break
+				}
+			}
+		}
+		false
+	}
+
 	pub fn new(record_id: u64) -> BTreeIterState {
```
Tpt, line 347: It's a bit strange to don't have `new` at the beginning of the `impl` block

Tpt (approved):
I don't feel knowledgeable enough to properly review this PR.

I just have some code style suggestions (but feel free to ignore it).

I have also fuzzed the code and it seems to work \o/.

cheme:
> I just have some code style suggestions (but feel free to ignore it).

they are very welcome

src/db.rs
cheme: I will keep it as is (the result is return as Vec<u8> since it can come from the file backend too, could return some Cow, but it will change all api so would be a bit overkill).

src/btree/iter.rs
cheme: yes, @i1i1 did write it this way too, I guess that's better, I did switch.
Warning I don't release the log lock anymore.

cheme:
@i1i1 if you got time to triple check this PR, review is very welcome (even if things get merge already I will create new PR).
At least please notice the change in iterator behavior (probably this will get published in semver major).

src/db.rs
Tpt: Ok! Make sense.

PR #110, opened 2022-09-15
Checks that preimage is enabled if ref_counted is

Tpt:


src/options.rs
```diff
@@ -84,6 +84,9 @@ impl ColumnOptions {
 			log::error!(target: "parity-db", "Using `preimage` option on an ordered column is not supported");
 			return false
 		}
+		if self.ref_counted && !self.preimage {
+			log::error!(target: "parity-db", "Using `ref_counted` option without `preimage` enabled is not supported");
```
cheme, line 88: I would return false (unless we just want the error logged).
Also I realize that the test above will block usage of ref-counted for btree, so I propose we simply remove the previous test.

cheme (approved):
LGTM.
So next version should be 0.4.0 (I was thinking of increasing minor for the iterator pr even if it changes iteration behavior, but seems better to follow strict versioning).
I wonder if we should have a changelog like we have in trie crate  https://github.com/paritytech/trie/blob/master/trie-db/CHANGELOG.md  cc @arkpar

cheme:
Oh the tests needs to be updated (I propose to set the preimage to true even if it is not really).

Tpt:
> Oh the tests needs to be updated (I propose to set the preimage to true even if it is not really).

Fixed

arkpar:
CHANGELOG for breaking changes is a good idea.
However we should eventually remove the `preimage` requirement for `ref_counted`.  Now that we have the `Operation` enum, `Set` should overwrite the value and reset the counter to 1.

PR #111, opened 2022-09-15
Replace blake2-rfc with blake2

davxy:
Currently Parity-DB is the only crate used by Substrate that is using `blake2-rfc`.

Everything else has been recently ported to RustCrypto `blake2`.

---

Please refer to this Substrate PR for some more details: https://github.com/paritytech/substrate/pull/12266

### DIFF
--- a/src/column.rs
+++ b/src/column.rs
@@ -125,6 +125,11 @@ enum IterStateOrCorrupted {
 
 #[inline]
 pub fn hash_key(key: &[u8], salt: &Salt, uniform: bool, db_version: u32) -> Key {
+	use blake2::{
+		digest::{typenum::U32, FixedOutput, Update},
+		Blake2bMac,
+	};
+
 	let mut k = Key::default();
 	if uniform {
 		if db_version <= 5 {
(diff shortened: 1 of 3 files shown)


src/column.rs
```diff
@@ -137,7 +142,11 @@ pub fn hash_key(key: &[u8], salt: &Salt, uniform: bool, db_version: u32) -> Key
 			}
 		}
 	} else {
-		k.copy_from_slice(blake2_rfc::blake2b::blake2b(32, salt, key).as_bytes());
+		let mut ctx = Blake2bMac::<U32>::new_with_salt_and_personal(salt, &[], &[])
+			.expect("Salt length (32) is a valid key length (<= 64)");
```
davxy, line 146: Here the naming is a bit misleading.

That is, with previous implementation the `salt` was used as the `blake2b-rfc` "key" argument while the `key` was used as the "data" argument.

The `new_with_salt_and_personal` has a dedicated parameter (the 2nd) for the salt. But to obtain the same MAC as before we have to pass `salt` as the 1st parameter

arkpar (approved):

cheme (approved):

PR #112, opened 2022-09-16
Improves fuzzers

Tpt:
Requires #109 to be fully functional but the code is independant from it.

### DIFF
--- a/fuzz/fuzz_targets/refcounted_model.rs
+++ b/fuzz/fuzz_targets/refcounted_model.rs
@@ -8,11 +8,10 @@
 use arbitrary::Arbitrary;
 use libfuzzer_sys::fuzz_target;
 use parity_db_fuzz::*;
-use std::collections::{
-	btree_map::{BTreeMap, Entry},
-	HashMap,
+use std::{
+	collections::{btree_map::Entry, BTreeMap, HashMap},
+	path::Path,
 };
-use tempfile::tempdir;
 
 #[derive(Arbitrary, Debug, Clone, Copy)]
 enum Operation {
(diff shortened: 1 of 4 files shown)


arkpar (approved):

fuzz/fuzz_targets/refcounted_model.rs
```diff
@@ -7,125 +7,77 @@
...
 				},
 		}
+	}
 
-		// We check the state
-		for (k, (_, v)) in &model {
-			assert_eq!(db.get(0, &[*k]).unwrap(), Some(vec![*v]));
+	fn map_operation(operation: &Operation) -> parity_db::Operation<Vec<u8>, Vec<u8>> {
+		match *operation {
+			Operation::Set(k) => parity_db::Operation::Set(vec![k], vec![k]),
```
cheme, line 69: We could use different length of keys and values here, either with smallrng or directly with fuzzer input (with fuzzer input should only make sense for btree).
But that's probably already on your roadmap.

fuzz/fuzz_targets/simple_model.rs
```diff
@@ -5,98 +5,59 @@
...
-
-			// We check the BTree backward iterator
-			model_content.reverse();
-			db_iter.seek_to_last().unwrap();
-			let mut db_content = Vec::new();
-			while let Some((k, v)) = db_iter.prev().unwrap() {
-				db_content.push((k[0], v[0]));
-			}
-			assert_eq!(db_content, model_content);
+	fn map_operation(operation: &(u8, Option<u8>)) -> parity_db::Operation<Vec<u8>, Vec<u8>> {
```
cheme, line 46: Could have an util function for sharing this logic with ref counted (for now it's not very useful actually:)

cheme (approved):
:+1:

fuzz/fuzz_targets/refcounted_model.rs
Tpt: Yes, definitely. It's on my roadmap. The tricky thing is that we want keys to get often reused between operations.

cheme: in my trie fuzzer I just for half the delete use a random index in the model

Tpt: Sounds like a great idea! I was also thinking about generating from the fuzzer input a set of keys and then picking in this set when generating operations.

PR #113, opened 2022-09-16
Allows to simulate I/O errors

Tpt:
Introduces the "instrumentation" feature.

When enabled the "set_number_of_allowed_io_operations" is exposed and allows to set the number of allowed I/O operations.
After this threshold is reached all I/O operations returns an error.

This required to wrap all I/O calls with the new "try_io!" macro that is an alias for "try!" if the "instrumentation" feature is not enabled.

### DIFF
--- a/src/index.rs
+++ b/src/index.rs
@@ -4,7 +4,7 @@
 use crate::{
 	column::ColId,
 	display::hex,
-	error::{Error, Result},
+	error::{try_io, Error, Result},
 	log::{LogQuery, LogReader, LogWriter},
 	stats::{self, ColumnStats},
 	table::{key::TableKey, SIZE_TIERS_BITS},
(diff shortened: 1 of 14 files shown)


arkpar:
Looks good, although this does not simulate mmap read/write failures. I think the correct way to simulate these would be to randomly send a SIGSEGV signal.  We can add this later though.

Tpt:
@arkpar I have updated the PR with a macro instead of the function.

It allows to fail *before* the I/O operation and, hence, get a better simulation.

src/index.rs
```diff
@@ -494,15 +494,15 @@ impl IndexTable {
 
 	pub fn drop_file(self) -> Result<()> {
 		drop(self.map);
```
cheme, line 496: I wonder if it would make sense to simulate munmap failure on drop (surely useless here on drop_file, but in general on db close). I mean error on flush will probably be flushed on mmap drop.

cheme (approved):

src/index.rs
Tpt: Yes, indeed. I have a question about it: would the drop panic or just silently fail?

cheme: I would think silently fail (no handling of error code in https://github.com/RazrFalcon/memmap2-rs/blob/64cf17da2b22fd6fbe6f86346b18906025487891/src/unix.rs#L284 ).

cheme: actually it is in the comment :facepalm: https://github.com/RazrFalcon/memmap2-rs/blob/64cf17da2b22fd6fbe6f86346b18906025487891/src/unix.rs#L279

PR #116, opened 2022-09-19
Makes metadata write more atomic-like

### LINKED ISSUE #115: DB corrupted on column config partial writes
When writing only a part of a column configuration, the DB gets corrupted:

To reproduce with #113:
```rust
	#[test]
	fn test_from_fuzzer() {
		crate::set_number_of_allowed_io_operations(4);
		let tmp = tempdir().unwrap();
		assert!(Db::open_or_create(&options).is_err()); //It's failing because of I/O limit
		Db::open_or_create(&options).unwrap(); //It's going to fail again with 'Column config mismatch'
	}
```

Tpt:
There is still two I/O operations (create + write_all) and write_all might also do multiple operations...

Closes #115

### DIFF
--- a/src/options.rs
+++ b/src/options.rs
@@ -7,7 +7,7 @@ use crate::{
 	error::{Error, Result},
 };
 use rand::Rng;
-use std::{collections::HashMap, io::Write, path::Path};
+use std::{collections::HashMap, path::Path};
 
 pub const CURRENT_VERSION: u32 = 7;
 // TODO on last supported 5, remove MULTIHEAD_V4 and MULTIPART_V4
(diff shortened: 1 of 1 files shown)


src/options.rs
```diff
@@ -175,13 +167,14 @@ impl Options {
...
-		writeln!(file, "salt={}", hex::encode(salt))?;
+		let mut metadata = vec![
+			format!("version={}", version.unwrap_or(CURRENT_VERSION)),
+			format!("salt={}", hex::encode(salt)),
+		];
 		for i in 0..self.columns.len() {
-			writeln!(file, "col{}={}", i, self.columns[i].as_string())?;
+			metadata.push(format!("col{}={}", i, self.columns[i].as_string()));
 		}
+		std::fs::write(path, metadata.join("\n"))?;
```
arkpar, line 177: Hmm, this is still not guaranteed to be atomic, is it? Looking at the documentaion, this calls `std::io::File::write` in a loop, which may result in a file written only partially. TBH I don't think this is likely to cause any issues in the real world. All of the metadata is likely to fit into a single 4kb IO page anyway.

arkpar (approved):

cheme (approved):

PR #123, opened 2022-09-21
Improves fuzzers

Tpt:
Values which reference number is 0 or below might be still present or not.

### DIFF
--- a/fuzz/fuzz_targets/refcounted_model.rs
+++ b/fuzz/fuzz_targets/refcounted_model.rs
@@ -52,10 +52,7 @@ impl DbSimulator for Simulator {
 			Operation::Dereference(k) =>
 				if let Entry::Occupied(mut e) = model.entry(k) {
 					let counter = e.get_mut();
-					*counter -= 1;
-					if *counter == 0 {
-						e.remove_entry();
-					}
+					*counter = counter.saturating_sub(1);
 				},
 			Operation::Reference(k) =>
 				if let Entry::Occupied(mut e) = model.entry(k) {
@@ -72,7 +69,14 @@ impl DbSimulator for Simulator {
 		}
 	}
 
-	fn model_content(model: &BTreeMap<u8, u8>) -> Vec<(Vec<u8>, Vec<u8>)> {
+	fn model_required_content(model: &BTreeMap<u8, u8>) -> Vec<(Vec<u8>, Vec<u8>)> {
+		model
+			.iter()
+			.filter_map(|(k, v)| if *v > 0 { Some((vec![*k], vec![*k])) } else { None })
+			.collect::<Vec<_>>()
+	}
+
+	fn model_optional_content(model: &BTreeMap<u8, u8>) -> Vec<(Vec<u8>, Vec<u8>)> {
 		model.iter().map(|(k, _)| (vec![*k], vec![*k])).collect::<Vec<_>>()
 	}
 }
--- a/fuzz/fuzz_targets/refcounted_model.rs
+++ b/fuzz/fuzz_targets/refcounted_model.rs
@@ -79,6 +79,10 @@ impl DbSimulator for Simulator {
 	fn model_optional_content(model: &BTreeMap<u8, u8>) -> Vec<(Vec<u8>, Vec<u8>)> {
 		model.iter().map(|(k, _)| (vec![*k], vec![*k])).collect::<Vec<_>>()
 	}
+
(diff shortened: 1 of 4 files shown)


arkpar (approved):

fuzz/fuzz_targets/refcounted_model.rs
```diff
@@ -52,10 +52,7 @@ impl DbSimulator for Simulator {
 			Operation::Dereference(k) =>
 				if let Entry::Occupied(mut e) = model.entry(k) {
 					let counter = e.get_mut();
-					*counter -= 1;
-					if *counter == 0 {
-						e.remove_entry();
-					}
+					*counter = counter.saturating_sub(1);
```
cheme, line 55: Reading how dereference work (in write_dec_ref), when reaching 0 next calls to Reference should be noops, so I am not sure why we did this change.

fuzz/src/lib.rs
```diff
@@ -45,7 +45,9 @@ pub trait DbSimulator {
 
 	fn map_operation(operation: &Self::Operation) -> parity_db::Operation<Vec<u8>, Vec<u8>>;
 
-	fn model_content(model: &Self::Model) -> Vec<(Vec<u8>, Vec<u8>)>;
+	fn model_required_content(model: &Self::Model) -> Vec<(Vec<u8>, Vec<u8>)>;
+
+	fn model_optional_content(model: &Self::Model) -> Vec<(Vec<u8>, Vec<u8>)>;
```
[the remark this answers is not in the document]
Tpt: Yes, definitely.  But it starts getting very quickly annoying because we are in a trait so `impl` return types don't work. It's why I have been lazy and I have not bothered to do it. It's likely to don't matter much anyway, we are opening/closing databases a lot during fuzzing so some extra allocations are likely to don't have any significant impact. Is it fine for you?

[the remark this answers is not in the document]
cheme: definitely fine :)

fuzz/src/lib.rs
```diff
@@ -150,13 +142,57 @@ pub trait DbSimulator {
...
+		db: &parity_db::Db,
+		model: &Self::Model,
+		is_db_b_tree: bool,
+	) -> parity_db::Result<Result<(), String>> {
+		for (k, v) in Self::model_required_content(model) {
+			if db.get(0, &k)?.as_ref() != Some(&v) {
+				return Ok(Err(format!("The value {:?} for key {:?} is not in the database", k, v)))
+			}
+		}
+
```
cheme, line 161: We should keep reference to removed content in model and explicitely check (when not ref_counted) that the removed content in model is not in db anymore.

fuzz/fuzz_targets/refcounted_model.rs
cheme: Oh, is it because the fuzzer might get its data from the commit overlay?
Ok I start to understand :facepalm: 
Indeed it's not trivial to know where the data is, so I get why we check in a less restrictive way.
I would think this 'optional model' less restrictive check should be only for ref counted and not in other case.

Tpt: This change is here because of two mistakes I made:
- Indeed, extra dereferences after reaching 0 are no op and not doing overflow as the current (i.e. before this PR) fuzzer model is doing.
- The database is not immediately garbage collecting the value when the counter reaches 0 so the value should still be stored in the model because database iteration might return it.

fuzz/src/lib.rs
Tpt: Done.

Tpt:
I have pushed also a change that fuzzes database initialization now that #122 is merged.

fuzz/fuzz_targets/refcounted_model.rs
cheme: Yes I did not get it at first, but make sense that we cannot do some very complete checks when using ref_counted (due to how things behave in the commit overlay).
Maybe using the optional model only when using ref_counted and keep an exhaustive check for btree without refcounted value can be a small improvment. (Simply having the optional model function return an optional value).

Tpt: >  Maybe using the optional model only when using ref_counted and keep an exhaustive check for btree without refcounted value can be a small improvment. (Simply having the optional model function return an optional value).

Yes, it's what is done now in PR after your comment about checking removed values. In `simple_model` removed keys are now checked for non-existence in the database.

cheme (approved):

PR #132, opened 2022-09-26
Inserts instrumented errors on mmap access

Tpt:
Allows to fuzz random disk crash

### DIFF
--- a/src/index.rs
+++ b/src/index.rs
@@ -208,24 +208,27 @@ impl IndexTable {
 		IndexTable { id, path, map: RwLock::new(None) }
 	}
 
-	pub fn load_stats(&self) -> ColumnStats {
+	pub fn load_stats(&self) -> Result<ColumnStats> {
 		if let Some(map) = &*self.map.read() {
-			ColumnStats::from_slice(&map[HEADER_SIZE..HEADER_SIZE + stats::TOTAL_SIZE])
+			Ok(ColumnStats::from_slice(try_io!(Ok(
+				&map[HEADER_SIZE..HEADER_SIZE + stats::TOTAL_SIZE]
+			))))
 		} else {
-			ColumnStats::empty()
+			Ok(ColumnStats::empty())
 		}
 	}
 
-	pub fn write_stats(&self, stats: &ColumnStats) {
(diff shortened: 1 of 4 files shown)


src/index.rs
```diff
@@ -208,24 +208,27 @@ impl IndexTable {
...
-			ColumnStats::empty()
+			Ok(ColumnStats::empty())
 		}
 	}
 
-	pub fn write_stats(&self, stats: &ColumnStats) {
+	pub fn write_stats(&self, stats: &ColumnStats) -> Result<()> {
 		if let Some(map) = &mut *self.map.write() {
-			let slice = &mut map[HEADER_SIZE..HEADER_SIZE + stats::TOTAL_SIZE];
+			let slice = try_io!(Ok(&mut map[HEADER_SIZE..HEADER_SIZE + stats::TOTAL_SIZE]));
```
cheme, line 223: if we want to fail when writing (or accessing) in the mmap, there is line 461 or 464
```
		log.read(&mut chunk[i as usize * ENTRY_BYTES..(i as usize + 1) * ENTRY_BYTES])?;
```
that is not cover ('read' here is a bit confusing as it actually fill a biffer that is a mmap and is indeed a write operation).

Tpt: Thank you! I have added it.

cheme: Nice. I don't know if it should be a rule in parity-db repo (I know it is the case for substrate and most paritytech repo), but could we avoid 'force-pushed'.
Not necessarily as a strict rule. If the PR did change significantly it can make sense to clean history, but for small change it is often easier to review when change history is easily accessible and we will squash and merge at the end. (it would just mean cleaning a bit the comment when doing squash and merge).

cheme (approved):

src/index.rs
cheme: Well I realize there is a `compare` button next to the force-push that seems to do the job so it is not necessarily an issue.

cheme: I think there was a few merge in the past but usually we want to squash merge, so I think it was mistake (button defaults to squash merge for me at least).

arkpar (approved):

PR #134, opened 2022-09-27
Uses tempdir in all tests

Tpt:
Not sure if it is welcome because it makes debugging a bit more cumbersome.


arkpar (approved):

cheme (approved):
I never did read the actual dir in the past so I am totally fine with this change.

PR #136, opened 2022-09-29
Removes skip_check_lock option

Tpt:
Not actually required anywhere


cheme (approved):
was use in https://github.com/paritytech/parity-db/pull/47/files#r983526644 (at the time there was error when reopening, but seems not needed anymore).
:+1:

PR #139, opened 2022-10-05
Recovery: only clear log on corruption

Tpt:
Avoids to drop all logs in case of e.g. an I/O error or a bug in the code.

It's already the behavior when opening the reader.


Tpt:
Bad idea: if the log file has been partially written we don't want to hard fail.

PR #140, opened 2022-10-05
Don't panic on bad size header

arkpar:


src/table.rs
```diff
@@ -463,6 +463,14 @@ impl ValueTable {
 					},
 				}
 			}
+
+			if buf.offset() > entry_end {
+				return Err(crate::error::Error::Corruption(format!(
+					"Unexpected entry size. Expected at least {} bytes",
+					entry_end
```
cheme, line 470:
```suggestion
					buf.offset() - 2
```
2 assume we only read size.

cheme (approved):
Agree with bound check, minor nit of the size in error message.

Tpt (approved):

PR #141, opened 2022-10-10
I/O simulation: always fail after the first failure

Tpt:
It simulates better a disk or network error that the previous behavior of only failing once.

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -65,8 +65,12 @@ pub fn set_number_of_allowed_io_operations(val: usize) {
 macro_rules! try_io {
 	($e:expr) => {{
 		if crate::error::IO_COUNTER_BEFORE_ERROR
-			.fetch_sub(1, ::std::sync::atomic::Ordering::Relaxed) ==
-			0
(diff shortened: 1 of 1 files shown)


arkpar (approved):

src/error.rs
```diff
@@ -65,8 +65,12 @@ pub fn set_number_of_allowed_io_operations(val: usize) {
 macro_rules! try_io {
 	($e:expr) => {{
 		if crate::error::IO_COUNTER_BEFORE_ERROR
-			.fetch_sub(1, ::std::sync::atomic::Ordering::Relaxed) ==
-			0
+			.fetch_update(
+				::std::sync::atomic::Ordering::SeqCst,
+				::std::sync::atomic::Ordering::SeqCst,
+				|v| Some(v.saturating_sub(1)),
```
cheme, line 71: Could also `|v| Some(v - 1)`. And compare to 1, but saturating and compare to 0 may be more readable. Looks good.

cheme (approved):

PR #143, opened 2022-10-10
Makes fuzzer call explicitly I/O operations

Tpt:
Makes fuzzing way more deterministic

Allows a more careful model of which state are already  persisted and which state is not

Does also some refactoring of `Db` test related API to be able to call them from the fuzzer

### DIFF
--- a/fuzz/fuzz_targets/refcounted_model.rs
+++ b/fuzz/fuzz_targets/refcounted_model.rs
@@ -21,10 +21,8 @@ enum Operation {
 
 #[derive(Clone, Debug)]
 struct Layer {
-	counts: [Option<usize>; NUMBER_OF_POSSIBLE_KEYS], /* The number of references per key (or
-	                                                   * None
-	                                                   * if the key never existed in the
-	                                                   * database) */
+	// The number of references per key (or None if the key never existed in the database)
+	counts: [Option<usize>; NUMBER_OF_POSSIBLE_KEYS],
 	written: bool,
 }
 
--- a/fuzz/fuzz_targets/refcounted_model.rs
+++ b/fuzz/fuzz_targets/refcounted_model.rs
@@ -10,7 +10,7 @@ use libfuzzer_sys::fuzz_target;
 use parity_db_fuzz::*;
 use std::{collections::HashMap, path::Path};
 
-const NUMBER_OF_POSSIBLE_VALUES: usize = 256;
+const NUMBER_OF_POSSIBLE_KEYS: usize = 256;
 
 #[derive(Arbitrary, Debug, Clone, Copy)]
 enum Operation {
@@ -21,8 +21,9 @@ enum Operation {
 
 #[derive(Clone, Debug)]
 struct Layer {
-	counts: [Option<i64>; NUMBER_OF_POSSIBLE_VALUES],
-	is_maybe_saved: bool,
+	counts: [Option<i64>; NUMBER_OF_POSSIBLE_KEYS], /* The number of references per key (or None
+	                                                 * if the key never existed in the database) */
+	written: bool,
 }
 
 type Model = Vec<Layer>;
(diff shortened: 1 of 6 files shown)


fuzz/fuzz_targets/simple_model.rs
```diff
@@ -7,17 +7,16 @@
...
-
-#[derive(Default)]
-struct Model {
-	current: BTreeMap<u8, u8>,
-	removed: BTreeSet<u8>,
+use std::{collections::HashMap, path::Path};
+
+#[derive(Clone, Debug)]
+struct Layer {
+	values: [Option<Option<u8>>; 256],
```
arkpar, line 14: Quite a few 255/256 values all over the code. Lets make it constant or use something like `model.values.len()` wherever possible, instead of hardcoding it.

fuzz/src/lib.rs
```diff
@@ -64,90 +79,77 @@ pub trait DbSimulator {
...
 						.unwrap();
+				},
+				Action::ProcessReindex =>
+					db = Self::try_or_restart(|db| db.process_reindex(), db, &mut model, &options),
+				Action::ProcessCommits => {
+					Self::write_first_layer_to_disk(&mut model);
+					db = Self::try_or_restart(|db| db.process_commits(), db, &mut model, &options)
+				},
+				Action::FlushAndEnactLogs => {
+					for _ in 0..2 {
```
arkpar, line 96: Could use a comment

Tpt: Done

fuzz/fuzz_targets/simple_model.rs
Tpt: Done

fuzz/src/lib.rs
Tpt: Side note: it might  be relevant to refactor a bit the log system to allow independant enact and flush

fuzz/fuzz_targets/refcounted_model.rs
```diff
@@ -20,11 +17,19 @@ enum Operation {
 	Reference(u8),
 }
 
+#[derive(Clone, Debug)]
+struct Layer {
+	counts: [Option<i64>; 256],
+	is_maybe_saved: bool,
```
cheme, line 25: `is_maybe_saved` is a bit confusing name for me, iiuc it indicates a layer has been written to disk, should we simply use `written`?

fuzz/src/lib.rs
cheme: @Tpt I feel like you have force push over your changes (or just github ui bugging).

src/db.rs
```diff
@@ -867,12 +867,10 @@ impl Db {
 		};
 		let flush_thread = if start_threads {
 			let flush_worker_db = db.clone();
-			let min_log_size =
-				if matches!(inner_options.commit_stages, EnableCommitPipelineStages::DbFile) {
-					0
-				} else {
-					MIN_LOG_SIZE_BYTES
-				};
+			#[cfg(test)]
```
cheme, line 870: is it needed for feature `instrumentation` too?

fuzz/src/lib.rs
Tpt: It's the correct place I think. I have just force pushed a slight code style fix on my last commit.

fuzz/fuzz_targets/refcounted_model.rs
```diff
@@ -41,28 +46,85 @@ impl DbSimulator for Simulator {
...
-			},
-			Operation::Dereference(k) =>
-				if let Entry::Occupied(mut e) = model.entry(k) {
-					let counter = e.get_mut();
-					*counter = counter.saturating_sub(1);
+	fn apply_operations_on_model<'a>(
+		operations: impl IntoIterator<Item = &'a Operation>,
+		model: &mut Model,
+	) {
+		let mut counts = [None; 256];
```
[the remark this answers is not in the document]
Tpt: It's 256 and not 255 (=`u8::MAX`) because it's the number of possible values in a `u8`. I might do `1 << u8::BITS` but I am not sure if it's more readable.

fuzz/src/lib.rs
cheme: Indeed, lot of my review comment were  done before https://github.com/paritytech/parity-db/pull/143/commits/832c889dd20ab1d89b633223470c3d2a81a0e8aa (not sure why I don't have a refresh view on github).

Tpt (commented):
Thank you for the review!

arkpar (approved):

Tpt (commented):
Thank you @cheme for the review!

I have fixed or reply to your comments.

I have also made simple_model validation of removed keys stricter (it was unnecessarily lenient).

cheme:
> I believe it is not always the case. What if the DB crashes between write and flush? We set is_written during write, not after flush.

This sentence explains a lot to me. I was expecting is_written to be set when we consider things flushed (since running without thread we should be able to set it at this point deterministically, may make sense as a next step). 

> Changing Some(0) to None would mean to assume that the GC is collecting all unused key-value at this step and I am not sure it's the case in the DB.

yes that was my initial assumption :)


```
    Commit 1: Set([0], [0])
    Commit 2: Increment([0])

If we are unsure that commit 2 has been flushed properly there is now way to know from the presence/abscence of key [0] if the reference count of key [0] is 1 or 2.
```
I guess your example was more on a decrement, but I see what you mean.

cheme (approved):

Tpt:
Thank you!

>> I believe it is not always the case. What if the DB crashes between write and flush? We set is_written during write, not after flush.
> This sentence explains a lot to me. I was expecting is_written to be set when we consider things flushed (since running without thread we should be able to set it at this point deterministically, may make sense as a next step).

Yes! I started to write something about it but I encountered some issue. I prefered to keep it as a next step.

Tpt:
@cheme And when there are also I/O errors it creates "fun" states like "flushed if it has been actually written" or "maybe flushed"...

cheme:
In theory these state should be cover by a layer upward or when restarting by the WAL commit not being complete (or crc failing).
But you're right, for fuzzing we should not assume everything is implemented/sequenced properly.

Issue #142, opened 2022-10-10
Support adding columns to existing DB


sandreim:
Some possible options I am thinking are:
- `add_column` API
- or automatically handle opening the DB with a configuration that has more columns than existing DB.

https://github.com/paritytech/polkadot/pull/6106

arkpar:
That should not be too complicated and since it will save a complex migration I'm ok with supporting it. 
`add_column` API is a more explicit, so I'm in favor of it.

PR #144, opened 2022-10-11
Support for adding new columns in existing db

sandreim:
Fixes https://github.com/paritytech/parity-db/issues/142

https://github.com/paritytech/polkadot/pull/6106

### DIFF
--- a/src/db.rs
+++ b/src/db.rs
@@ -1030,10 +1030,16 @@ impl Db {
 
 	/// Add a new column with options specified by `new_column_options`.
 	pub fn add_column(options: &mut Options, new_column_options: ColumnOptions) -> Result<()> {
+		// We open the DB before to check metadata validity and make sure there are no pending WAL
+		// logs.
+		let db = Db::open(&options)?;
+		let salt = db.inner.options.salt;
+		drop(db);
+
 		options.columns.push(new_column_options);
 		options.write_metadata_with_version(
 			&options.path,
-			&options.salt.expect("Salt is required to add column"),
+			&salt.expect("`salt` is always `Some` after opening the DB; qed"),
 			Some(CURRENT_VERSION),
 		)?;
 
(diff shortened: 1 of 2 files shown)


src/db.rs
```diff
@@ -1027,6 +1027,18 @@ impl Db {
 	pub fn stats(&self) -> StatSummary {
 		self.inner.stats()
 	}
+
+	/// Add a new column with options specified by `new_column_options`.
+	pub fn add_column(options: &mut Options, new_column_options: ColumnOptions) -> Result<()> {
```
cheme, line 1032:
```suggestion
	pub fn add_column(options: &mut Options, new_column_options: ColumnOptions) -> Result<()> {
		options.load_and_validate_metadata(false)?;
```
Just to ensure we don't overwrite other column contents.

src/db.rs
```diff
@@ -1027,6 +1027,18 @@ impl Db {
...
+		options.columns.push(new_column_options);
+		options.write_metadata_with_version(
+			&options.path,
+			&options.salt.expect("Salt is required to add column"),
```
arkpar, line 1036: Should not be required really. When opening the database `None` means that existing value must be used. After you open the database as per previous comment, you can get the existing salt value.

src/db.rs
arkpar: It would be even better to make sure there are no pending WAL logs first. This can be done by opening and then closing the database.

sandreim: done!

src/db.rs
sandreim: done!

arkpar (approved):

cheme (approved):

src/db.rs
sandreim: I’ll note that in a comment as well.

arkpar:
Released in parity-db v0.4.2

PR #146, opened 2022-10-14
Clean all logs after recovery

### LINKED ISSUE #145: After recovery on corrupted log, existing values might get overriden after unrelated commit processing
Senario:
1. The database is created with reference-counting.
2. A first commit C1 is done, inserting a key K1, the commit is processed and the log is flushed
3. A first commit C2 is done, inserting a key K2 is inserted, and the commit is processed.
4. A third commit C3 is done (might be empty), but the I/Os fail during commit processing.
5. The database is closed and open again. The second log file is invalid and removed.
6. The database is closed and open a second time. K1 is still present in the DB.
7. A fourth commit C4 is done and processed. When writing a `Key mismatch at 2` log is printed.
8. K1 is not in the database anymore.

As a test:
```rust
	#[test]
	fn test_from_fuzzer() {
		let tmp = tempdir().unwrap();
		options.columns[0].ref_counted = true;
		options.columns[0].preimage = true;
		options.always_flush = true;
		options.with_background_thread = false;

		{
			db.commit_changes(vec![(0, Operation::Set(vec![59], vec![59]))]).unwrap();
			db.process_commits().unwrap();
			db.flush_logs().unwrap();
			db.commit_changes(vec![(0, Operation::Set(vec![0], vec![0]))]).unwrap();
			db.process_commits().unwrap();
			db.commit_changes(vec![]).unwrap();
			crate::set_number_of_allowed_io_operations(3);
			assert!(db.process_commits().is_err()); // Partial write
			crate::set_number_of_allowed_io_operations(usize::MAX);
		}

		{
			let db = Db::open(&options).unwrap();
			assert!(db.get(0, &[0]).unwrap().is_some()) // Key 0 is here
		}

		{
			let db = Db::open(&options).unwrap();
			assert!(db.get(0, &[0]).unwrap().is_some()); // Key 0 is here
			db.commit_changes(vec![(0, Operation::Set(vec![196], vec![196]))])
			.unwrap();
			db.process_commits().unwrap();
			assert!(db.get(0, &[0]).unwrap().is_none()) // Key 0 has been removed
		}
	}
```

Before C4 the database content is:
```
Index key: [22, 1f, f2, 5f, 13, 93, be, df, 22, cc, fd, 8e, 6b, 6b, bf, e1, bf, d6, 3f, 3c, 27, 2a, 9c, 1f, db, 53, be, a3, 18, 15, 32, e0]
        Rc: 1
Value: 3b
Index key: [f3, 71, 9b, 9e, 9f, be, 4, e3, 5f, 25, 5c, d7, a9, 29, c5, 24, d9, 8d, b1, 21, 19, 2f, 64, 6e, 52, c3, 3b, 9b, 1b, d8, 6d, 6f]
        Rc: 1
Value: 00
```

And after:
```
for 65536 chunks of column 0
Index key: [22, 1f, f2, 5f, 13, 93, be, df, 22, cc, fd, 8e, 6b, 6b, bf, e1, bf, d6, 3f, 3c, 27, 2a, 9c, 1f, db, 53, be, a3, 18, 15, 32, e0]
        Rc: 1
parity_db::column] Value: 3b
Index key: [a5, fa, ce, bb, 91, 75, b7, d7, c5, 4b, de, f6, 65, e1, a5, 2d, 82, b2, 29, 33, 2c, 60, ab, 29, ab, 57, 8, 15, b2, dc, c6, ed]
        Rc: 1
Value: c4
Index key: [f3, 71, 9b, 9e, 9f, be, b7, d7, c5, 4b, de, f6, 65, e1, a5, 2d, 82, b2, 29, 33, 2c, 60, ab, 29, ab, 57, 8, 15, b2, dc, c6, ed]
        Rc: 1
Value: c4
```

It seems like somehow the (0, 0) key-value partially overridden by the (196-196) key-value.

Tpt:
Allows to make sure no old log is staying after proper recovery

Fixes #145

### DIFF
--- a/src/db.rs
+++ b/src/db.rs
@@ -723,6 +723,9 @@ impl DbInner {
 			c.refresh_metadata()?;
 		}
 		log::debug!(target: "parity-db", "Replay is complete.");
+		log::debug!(target: "parity-db", "Cleaning logs.");
+		self.clean_all_logs()?;
+		log::debug!(target: "parity-db", "Log cleanup completed.");
 		Ok(())
 	}
 


Tpt:
I am not sure it's a clean fix of #145. It prevents more a cause of the bug than fixing it really.

cheme (approved):

arkpar:
Could you also add a test from #145 please.

The issue is indeed that the old log was not removed and was replayed, screwing the header on the value table. There's a [TODO](https://github.com/paritytech/parity-db/blob/master/src/table.rs#L881) to detect this issue at validation, which should have prevented it as well, once implemented. 

When validating, logs are removed out of order. Each successfully validated and enacted log is added to the cleanup queue. But if there's an error, the erroneous  log is dropped immediately and the cleanup queue is not processed at all.

While this fix does indeed cleans up everything, it is still in the wrong order. So if an (unlikely) crash happens when the logs are being cleaned, the bug would still be there.

Tpt:
@arkpar Thank you for your detailed explanation. So, I understand correctly the fix to do is to implement the [TODO](https://github.com/paritytech/parity-db/blob/master/src/table.rs#L881) to detect this issue at validation.

And, maybe, process the cleanup queue in case of a validation error. But if we have the validation then it should not be required anymore if I understand correctly.

I am closing this PR and I'm going to write an other one.

arkpar:
Fixing the TODO wont be enough in the general case. If you don't change the number of value table entries, but just modify an existing value, you'll end up with a slightly different version of this bug. The log would be valid, but the value will be old. So I would not bother with it for now.

I suppose the most relevant fix would be to

1. Not drop the log in `clear_replay_log`, but instead add it to the `cleanup_queue`
2. Process the cleanup queue even the database is shut down with error in `kill_logs`.

Tpt:
@arkpar Thank you!

PR #149, opened 2022-10-18
Makes fuzzer support more recovery edge cases

Tpt:
* Makes fuzzer support the recovery case "the disk becames unreadable during recovery so the fuzzer model believes the database is empty. After the subsequence DB restart the recovery process works and past commits get properly reneacted, leading to a DB state different from the fuzzer model".
* Removes a lot of code duplication in the fuzzers


Tpt:
This change is much easier to review commit by commit rater than reading the big final diff.

arkpar (approved):

PR #151, opened 2022-10-20
Log: Adds a read queue

Tpt:
Allows to avoid having a global state on Log (Reading vs Idle) and to be able to flush log even if the next file is not fully enacted

### DIFF
--- a/src/log.rs
+++ b/src/log.rs
@@ -755,7 +755,7 @@ impl Log {
 		&self.overlays
 	}
 
-	pub fn has_still_log_files_to_read(&self) -> bool {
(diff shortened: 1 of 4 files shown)


arkpar:
This looks like a nice simplification

src/log.rs
```diff
@@ -815,6 +755,10 @@ impl Log {
 		&self.overlays
 	}
 
+	pub fn has_still_log_files_to_read(&self) -> bool {
```
arkpar, line 758: `still_has_log_files_to_read` or simply `has_log_files_to_read`

Tpt: Done!

PR #154, opened 2022-10-24
Properly reset BTree iteration state on boundaries

### LINKED ISSUE #152: Iteration behavior changes depending on the log worker thread work
In the following test, the presence or abscence of the `db.process_commits` call changes the iteration behavior of the last `iter.prev()` call.

If `db.process_commit` is never called, the last `iter.prev()` returns the key `[0]` as exepected, if `db.process_commit` is called before iterating, it returns `None`.

This test is doing both forward and backward iterations, I don't know if we want to support both on the same iterator...

```rust
	#[cfg(feature = "instrumentation")]
	#[test]
	fn test_process_commit_in_the_middle_of_iter() {
		let tmp = tempdir().unwrap();
		options.always_flush = true;
		options.with_background_thread = false;

		db.commit::<_, Vec<u8>>(vec![(0, vec![0], Some(vec![0]))]).unwrap();
		db.process_commits().unwrap();
		let mut iter = db.iter(0).unwrap();
		iter.seek_to_last().unwrap();
		assert_eq!(iter.prev().unwrap(), Some((vec![0], vec![0]))); // State = 0
		assert_eq!(iter.next().unwrap(), None); // State = End
		assert_eq!(iter.prev().unwrap(), Some((vec![0], vec![0]))); // Fails because is None, work properly if process_commit is not called
	}
```

Changelog: 2022-10-24: Updated the bug description, the bug happens even if `process_commits` is called before iterating.

Tpt:
Closes #152

### DIFF
--- a/src/btree/iter.rs
+++ b/src/btree/iter.rs
@@ -277,7 +277,7 @@ impl BTreeIterState {
 
 	fn exit(&mut self, direction: IterDirection) -> bool {
 		loop {
-			if self.state.len() < 2 {
+			if self.state.is_empty() {
 				return true
 			}
 			self.state.pop();
--- a/src/btree/iter.rs
+++ b/src/btree/iter.rs
@@ -276,10 +276,7 @@ impl BTreeIterState {
 	}
 
 	fn exit(&mut self, direction: IterDirection) -> bool {
-		loop {
-			if self.state.is_empty() {
-				return true
-			}
+		while !self.state.is_empty() {
 			self.state.pop();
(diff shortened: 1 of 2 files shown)


cheme (approved):
LGTM, an alternate implementation would have been to keep the last node when exiting and and change the state to avoid fetching root again (probably before(index +1) or before(index) depending on direction), but considering this is a small corner case and already a bug source, I think this solution is better.

Tpt:
@cheme Thank you!

PR #157, opened 2022-10-26
Allows to run the fuzzers using HonggFuzz

Tpt:


Tpt:
Not sure we want to merge this. I have not been able to find new bugs running honggfuzz insead of libfuzzer

cheme (approved):
I like having both option, even if not strictly necessary. So I would be pro merge, but no strong opinion here.

arkpar:
Unused features tend to be unmaintained and become broken. I'd prefer to not introduce any extra maintenance unless it has any advantages over libfuzzer

PR #161, opened 2022-10-31
Avoid a race-condition leading to crash in Log::clean_logs

Tpt:
If Log.cleanup_queue is partially drained between the computation of the number of logs to clean (the "count" parameter) and the actual call to "drain", the drain method was panicking.

The changed code only drains at most the number of elements of the queue, avoiding the panic risk.

Found using Loom


arkpar (approved):

cheme (approved):
lgtm (we are avoiding the crash I think)

PR #163, opened 2022-11-14
Tests writes and iteration with Loom

Tpt:


.github/workflows/rust.yml
```diff
@@ -79,3 +79,5 @@ jobs:
         with:
           command: test
           args: --test loom --features=loom,instrumentation --release --verbose
+          env:
+            LOOM_MAX_PREEMPTIONS: 3
```
Tpt, line 83: We limit here the number of context switch to keep the number of sequence of states that Loom checks to an amount compatible with a usable CI runtime.

src/table.rs
```diff
@@ -155,6 +155,9 @@ impl Header {
 }
 
 pub struct Entry<B: AsRef<[u8]> + AsMut<[u8]>>(usize, B);
+#[cfg(feature = "loom")]
+pub type FullEntry = Entry<Vec<u8>>;
```
Tpt, line 159: Bad hack because of the stack limitations introduced by Loom.

arkpar:
Is it possible to increase the stack size with `ulimit` on the CI runner instead of changing the code?

Tpt:
> Is it possible to increase the stack size with `ulimit` on the CI runner instead of changing the code?

I tried to play with it and with `RUST_MIN_STACK` that configures the spawned threads limit. Sadly I have not manage to get the tests to pass without either crashing or consuming all my laptop memory (32GB). It's quite unfortunate.

arkpar (approved):

tests/loom.rs
```diff
@@ -6,31 +6,100 @@ use std::sync::Arc;
...
+			db.commit::<_, Vec<u8>>(vec![(0, vec![0], Some(vec![3])), (0, vec![1], Some(vec![3]))])
+				.unwrap();
+			db.enact_logs().unwrap();
 		});
 
 		t1.join().unwrap();
 		t2.join().unwrap();
+		t3.join().unwrap();
 
 		// Checks, that either T1 committed before T2 or reverse
```
cheme, line 56: So IIUC here the assert_eq is not really what we want to test, comment may not be up to date. Would just put `Check state consistency`.

cheme (approved):

PR #164, opened 2022-11-16
Fixes CI workflow file syntax

Tpt:


Tpt:
Sorry for that... It seems that GitHub does not display any warning on PRs if the worflow file is broken...

PR #166, opened 2022-12-12
Update README.md

b0xtch:


Tpt (approved):
Thank you! We have a "two reviewers" policy so we need to wait for an other approval before merge.

cheme (approved):

PR #167, opened 2022-12-14
implement FromStr for CompressionType

LesnyRumcajs:
Added `FromStr` for `CompressionType`

This is useful for clients with compression type adjustable via, e.g., a config file or CLI flag.


Tpt (approved):
Thank you! It looks great to me! We have a "two reviewers" policy so we need to wait for an other approval before merge.

PR #172, opened 2023-01-30
Makes Clippy happy

Tpt:


Tpt:
The "[Uses inline variable names in fmt strings](https://github.com/paritytech/parity-db/commit/e9777bce0c49683a8e51af0456f1aa66e9df337c)" commit has been done automatically with `cargo clippy --fix`

src/column.rs
```diff
@@ -409,10 +409,10 @@ impl HashColumn {
 		self.write_reindex_plan_locked(tables, reindex, key, address, log)
 	}
 
-	fn write_reindex_plan_locked<'a, 'b>(
+	fn write_reindex_plan_locked(
 		&self,
-		mut tables: RwLockUpgradableReadGuard<'a, Tables>,
-		mut reindex: RwLockUpgradableReadGuard<'b, Reindex>,
+		mut tables: RwLockUpgradableReadGuard<'_, Tables>,
+		mut reindex: RwLockUpgradableReadGuard<'_, Reindex>,
```
cheme, line 415: Can they be just ignored? (or somehow having two _ indicate they are different)

cheme (approved):

src/column.rs
Tpt: Yes, they can. I have pushed an extra commit in this PR removing them (and some others)

cheme: Thanks :)

PR #175, opened 2023-01-30
IndexTable::find_entry: add assertion bound checking

Tpt:
Allows to avoid a bound check on each `read_entry` call


arkpar (approved):

cheme (approved):
Did not see the linked issue the other day, makes sense. Probably not needed for  `plan_insert_chunk` (buffer size known).

Maybe a comment would be good, no comment alternative would be to change chunk parameter to &[u8; CHUNK_SIZE] and unsafe cast before calling find_entry.

PR #176, opened 2023-01-31
Use SSE2 SIMD in IndexTable::find_entry

Tpt:


src/index.rs
```diff
@@ -231,8 +235,63 @@ impl IndexTable {
...
+		#[cfg(any(target_arch = "x86", target_arch = "x86_64"))]
+		if is_x86_feature_detected!("sse2") {
+			return self.find_entry_sse2(key_prefix, sub_index, chunk)
+		}
+		self.find_entry_regular(key_prefix, sub_index, chunk)
+	}
+
+	#[cfg(any(target_arch = "x86", target_arch = "x86_64"))]
+	fn find_entry_sse2(&self, key_prefix: u64, sub_index: usize, chunk: &[u8]) -> (Entry, usize) {
+		debug_assert!(
```
arkpar, line 248: `sub_index` is the starting index for the search, rather than the number of entities. There may be multiple entries mathing the partial key in a single chunk. `find_entry` is called in a loop and all matching entries are further validated for the full key that's stored in the value table. `sub_index` is set to the entry index of the previous match, so that the search continues if the previous match did not pass that validation. Therefore it can be arbitatrary. We can just round it here to the prevous multiple of four. E.g. if `sub_index` is 7, set it to 4 and skip the first 3 result bits. 

I think we can also use aligned loads in this case. Though this will probably require changing the type from `IndexChunk` to a struct with alignment attribute here:
https://github.com/paritytech/parity-db/blob/master/src/log.rs#L444

src/index.rs
```diff
@@ -231,8 +235,63 @@ impl IndexTable {
...
+				let last_two = _mm_shuffle_epi32::<0b11011000>(_mm_srl_epi64(
+					_mm_loadu_si128(chunk[(i + 2) * 8..].as_ptr() as *const __m128i),
+					_mm_set_epi64x(0, shift.into()),
+				));
+				// We set into current the low parts of first_two and last_two
+				// in the interleaved order
+				let current = _mm_unpacklo_epi32(first_two, last_two);
+				let cmp = _mm_movemask_epi8(_mm_cmpeq_epi32(current, target));
+				if cmp != 0 {
+					// We found a match, let's validate it
```
arkpar, line 274: I think here we can simply return entry at `i + cmp.leading_zeroes() / 4`

arkpar:
Great stuff. Curios to see if it has any affect on the query performance.

src/index.rs
```diff
@@ -231,8 +235,63 @@ impl IndexTable {
...
+						}
+					}
+				}
+				i += 4;
+			}
+		}
+		(Entry::empty(), 0)
+	}
+
+	#[inline(never)]
```
arkpar, line 288: This was added by accident in #170 and should be removed.

src/index.rs
```diff
@@ -231,8 +235,63 @@ impl IndexTable {
...
+
+		unsafe {
+			let partial_key = Entry::extract_key(key_prefix, self.id.index_bits());
+			let target = _mm_set1_epi32(partial_key as i32);
+			let shift = Entry::address_bits(self.id.index_bits());
+			let mut i = sub_index;
+			while i + 4 <= CHUNK_ENTRIES {
+				// We load the value two by two and do the same shift as for the partial_key
+				// before moving the values into the low part of the register
+				let first_two = _mm_shuffle_epi32::<0b11011000>(_mm_srl_epi64(
```
cheme, line 261: I am a bit lost in my reading, I would get 1 and 3 at first then both get interleaves, so would expect 0b0111...
(before shift the partial is pos 0 and 2 and then 1 and 3).
But I might mess my ordering here.

src/index.rs
Tpt: The tricky thing is that the shift might be of 30 bits i.e. are left with a 34 bit number that we cast to 32 bits. So, this means we need to check again, sadly. But we can definitely be smarter and skip not useful comparison.

cheme: Isn't the idea to just skip two bits of check in this case? (but then need to shift 32 bits instead of 30, so we check the 32 bit of the start, the 2 last unchecked bits will be checked with the rest of the key on value access).

cheme: Spend time remembering/reunderstanding this today. might still be wrong, but my conclusion was the higher 2 so we in case of index 16 we would need to move >> 32 instead of 30. (the other way we would miss checking key content as the part of the key stored and check in index is 26 bytes containing the last two bits, then we indeed got 16 bit check from the access to the index, 32 from the index key content and 26 * 8 from the value table, should be 256).

Tpt: Thank you! I'm going to update the PR to use the high 32 bits.

Tpt: Done

src/index.rs
Tpt: Done

src/index.rs
Tpt: Done.

Tpt:
@arkpar @cheme Thank you for  your reviews. I have written a version 2 using the assumption that the high 32bits are part of the partial key.

I ran a quick benchmark on my laptop. The SSE2 version seems to be slightly faster but the numbers are not stable enough to tell more than that. We might also try an AVX version.

src/index.rs
arkpar: Previous version was actually more correct. We can't assume that partial key is always 32 bits, so shifts are still needed. We can ignore the last two bits for the cases when it is 33 or 34 bytes however. A bit of explanation:

Keys are always 256 bits. Key bits [48..256] are stored in the value table. So if `find_entry` returns a match, these bits are compared to make sure the rest of the key matches . Therefore `find_entry` must only match the first 48 bits of the key. Current version actually matches 50, so that's a bit of an overkill. 

The first `index_bits` of the key are simply the offset of the chunk in the file. All entries in the chunk share the same first `index_bits` of the key. They are not stored in the entries. Bits [index_bits..50] is what's actually stored in the entry and that's called `partial_key`.  The rest of the entry bits is `address`. As `index_bits` grows, so does `address_bits`, and `partial_key` shrinks. But the chunk index bits grows as well, so we can always match 50 bits of the key in the index file.

Indexes start with  `index_bits==16`. With  `index_bits==16` `partial_key` is 34, but the last two bits can be ignored and we only need to match 48 bits.
With `index_bits==21` however  `partial_key` is only 29 bits, so the last 3 of the first 32 entry bits should be excluded from comparison.

src/index.rs
```diff
@@ -231,8 +235,70 @@ impl IndexTable {
...
+			while i + 4 <= CHUNK_ENTRIES {
+				// We load the value 2 by 2 and move the high bits into the low part of the register
+				let first_two = _mm_shuffle_epi32::<0b10001101>(_mm_loadu_si128(
+					chunk[i * 8..].as_ptr() as *const __m128i,
+				));
+				let last_two = _mm_shuffle_epi32::<0b10001101>(_mm_loadu_si128(
+					chunk[(i + 2) * 8..].as_ptr() as *const __m128i,
+				));
+				// We set into current the input low parts in the interleaved order
+				let current = _mm_unpacklo_epi32(first_two, last_two);
```
arkpar, line 271: Maybe if we use `_mm_unpacklo_epi64` instead, the parts effectively would not be interleaved and it should be possible to use `leading_zeros()` with the mask?

src/index.rs
```diff
@@ -231,8 +235,70 @@ impl IndexTable {
 		Ok(try_io!(Ok(&map[offset..offset + CHUNK_LEN])))
 	}
 
-	#[inline(never)]
 	fn find_entry(&self, key_prefix: u64, sub_index: usize, chunk: &[u8]) -> (Entry, usize) {
+		#[cfg(any(target_arch = "x86", target_arch = "x86_64"))]
+		if is_x86_feature_detected!("sse2") {
```
arkpar, line 240: I'd use compile time rather than runtime detection. Similar to how's it [done in `HashMap` ](https://github.com/rust-lang/hashbrown/blob/master/src/raw/mod.rs#L22). I mean we use `HashMap` a lot in this code, and if it breaks at runtime because `target_arch` was incorrect, we're screwed anyway.

src/index.rs
```diff
@@ -231,8 +235,70 @@ impl IndexTable {
...
+		assert!(chunk.len() >= CHUNK_ENTRIES * 8); // Bound checking (not done by SIMD instructions)
+		debug_assert!(
+			Entry::address_bits(self.id.index_bits()) <= 32,
+			"To be sure we can use all high 32 bits as key prefix"
+		);
+		debug_assert_eq!(
+			CHUNK_ENTRIES % 4,
+			0,
+			"We assume here we got buffer with a number of elements that is a multiple of 4"
+		);
```
arkpar, line 257: Could be a compile time constant
```suggestion
		const _: () = assert_eq!(
			CHUNK_ENTRIES % 4,
			0,
			"We assume here we got buffer with a number of elements that is a multiple of 4"
		);
```

Tpt: Done

src/index.rs
Tpt: Done. Indeed.

src/index.rs
Tpt: Thank you! That's much better! Done.

src/index.rs
```diff
@@ -231,7 +235,46 @@ impl IndexTable {
...
 
-	#[inline(never)]
+	#[cfg(target_feature = "sse2")]
+	fn find_entry(&self, key_prefix: u64, sub_index: usize, chunk: &[u8]) -> (Entry, usize) {
+		assert!(chunk.len() >= CHUNK_ENTRIES * 8); // Bound checking (not done by SIMD instructions)
+		debug_assert!(
+			Entry::address_bits(self.id.index_bits()) <= 32,
+			"To be sure we can use all high 32 bits as key prefix"
+		);
+		const _: () = assert!(
```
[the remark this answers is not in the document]
arkpar: There used to be such a macro in parity-db at some point but we've removed it after rustc added support for panics in const expressions. `const _: () = assert!(..)` is simple enough.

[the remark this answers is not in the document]
cheme: I was on an old branch, thought it was still here.

src/index.rs
Tpt: Thank you for your detailed explanation. I indeed forgot that the index_bit can grow further. I have applied @cheme comments to fix the implementation.

src/index.rs
```diff
@@ -231,7 +235,46 @@ impl IndexTable {
...
+			CHUNK_ENTRIES % 4 == 0,
+			"We assume here we got buffer with a number of elements that is a multiple of 4"
+		);
+
+		unsafe {
+			let target = _mm_set1_epi32(((key_prefix << self.id.index_bits()) >> 32) as i32);
```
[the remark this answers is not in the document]
Tpt: Done. I wrote the code a bit differently to hilight that we want to shift by at least 32 bits to fit properly into register.

src/index.rs
```diff
@@ -231,7 +235,46 @@ impl IndexTable {
...
+			let mut i = (sub_index >> 2) << 2; // We keep an alignment of 4
+			while i + 4 <= CHUNK_ENTRIES {
+				// We load the value 2 by 2 and move the high bits into the low part of the register
```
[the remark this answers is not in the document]
Tpt: Done. Thank you for the code.

src/index.rs
```diff
@@ -231,7 +235,46 @@ impl IndexTable {
...
+						return (Self::read_entry(chunk, position), position)
+					}
+				}
+				i += 4;
+			}
+		}
+		(Entry::empty(), 0)
+	}
+
+	#[cfg(not(target_feature = "sse2"))]
```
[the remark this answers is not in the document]
Tpt: I have added a test that covers explicitely both implementations.

arkpar (approved):

cheme (approved):
minor nit, could do :
```
diff --git a/src/index.rs b/src/index.rs
index 5bd78b9..f5e94bf 100644
--- a/src/index.rs
+++ b/src/index.rs
@@ -235,12 +235,14 @@ impl IndexTable {
                Ok(try_io!(Ok(&map[offset..offset + CHUNK_LEN])))
        }
 
+       #[cfg(target_feature = "sse2")]
        fn find_entry(&self, key_prefix: u64, sub_index: usize, chunk: &[u8]) -> (Entry, usize) {
-               if cfg!(target_feature = "sse2") {
-                       self.find_entry_sse2(key_prefix, sub_index, chunk)
-               } else {
-                       self.find_entry_base(key_prefix, sub_index, chunk)
-               }
+               self.find_entry_sse2(key_prefix, sub_index, chunk)
+       }
+
+       #[cfg(not(target_feature = "sse2"))]
+       fn find_entry(&self, key_prefix: u64, sub_index: usize, chunk: &[u8]) -> (Entry, usize) {
+               self.find_entry_base(key_prefix, sub_index, chunk)
        }
 
        #[cfg(target_feature = "sse2")]
@@ -284,6 +286,7 @@ impl IndexTable {
                (Entry::empty(), 0)
        }
 
+       #[cfg(any(not(target_feature = "sse2"), test))]
        fn find_entry_base(&self, key_prefix: u64, sub_index: usize, chunk: &[u8]) -> (Entry, usize) {
                assert!(chunk.len() >= CHUNK_ENTRIES * 8);
                let partial_key = Entry::extract_key(key_prefix, self.id.index_bits());
@@ -608,19 +611,19 @@ mod test {
 
        #[test]
        fn test_find_entries() {
```
or expose a dummy `find_entry_sse2` function when no feature sse2 (method may be missing if no traget).

PR #178, opened 2023-02-03
Reduce alloc in copy_to_overlay

hanabi1224:
I notice that the `copy_to_overlay` method clones the `value` vector in Hash index mode and clones both the `key` and `value` vectors in Btree index mode, which seems to be an unnecessary overhead, this PR tries to, instead, clone the pointer type `Arc<Vec<u8>>` which is much cheaper, please let me know your feedback, thanks!

### DIFF
--- a/src/btree/iter.rs
+++ b/src/btree/iter.rs
@@ -57,8 +57,8 @@ struct PendingBackend {
 pub enum LastKey {
 	Start,
 	End,
-	At(RcValue),
-	Seeked(RcValue),
+	At(Vec<u8>),
+	Seeked(Vec<u8>),
 }
 
 #[derive(Debug)]
@@ -161,7 +161,7 @@ impl<'a> BTreeIterator<'a> {
 							if let Some(value) = commit_value {
 								Some((commit_key, value))
 							} else {
-								self.last_key = LastKey::At(commit_key);
+								self.last_key = LastKey::At(commit_key.value().clone());
 								continue
 							}
 						},
@@ -172,7 +172,7 @@ impl<'a> BTreeIterator<'a> {
 							if let Some(value) = commit_value {
 								Some((backend_key.into(), value))
 							} else {
-								self.last_key = LastKey::At(commit_key);
+								self.last_key = LastKey::At(commit_key.value().clone());
 								continue
 							},
 					}
@@ -183,7 +183,7 @@ impl<'a> BTreeIterator<'a> {
 				},
 				(Some((k, None)), None) => {
 					self.pending_backend = Some(PendingBackend { next_item: None, direction });
-					self.last_key = LastKey::At(k);
+					self.last_key = LastKey::At(k.value().clone());
 					continue
 				},
 				(None, Some((backend_key, backend_value))) =>
@@ -196,7 +196,7 @@ impl<'a> BTreeIterator<'a> {
 
 			match result.as_ref() {
 				Some((key, _)) => {
-					self.last_key = LastKey::At(key.clone());
+					self.last_key = LastKey::At(key.value().clone());
 				},
 				None =>
 					self.last_key = match direction {
@@ -221,10 +221,10 @@ impl<'a> BTreeIterator<'a> {
 			*tree = new_tree;
 			match &self.last_key {
 				LastKey::At(last_key) => {
-					iter.seek(SeekTo::Exclude(last_key.value().as_slice()), tree, col, log)?;
+					iter.seek(SeekTo::Exclude(last_key.as_slice()), tree, col, log)?;
 				},
 				LastKey::Seeked(last_key) => {
-					iter.seek(SeekTo::Include(last_key.value().as_slice()), tree, col, log)?;
+					iter.seek(SeekTo::Include(last_key.as_slice()), tree, col, log)?;
 				},
 				LastKey::Start => {
 					iter.seek(SeekTo::Include(&[]), tree, col, log)?;
--- a/src/btree/iter.rs
+++ b/src/btree/iter.rs
@@ -152,8 +152,8 @@ impl<'a> BTreeIterator<'a> {
 			let result = match (next_commit_overlay, next_backend) {
 				(Some((commit_key, commit_value)), Some((backend_key, backend_value))) => {
 					match (direction, commit_key.value().cmp(&backend_key)) {
-						(IterDirection::Backward, std::cmp::Ordering::Greater)
-						| (IterDirection::Forward, std::cmp::Ordering::Less) => {
+						(IterDirection::Backward, std::cmp::Ordering::Greater) |
+						(IterDirection::Forward, std::cmp::Ordering::Less) => {
 							self.pending_backend = Some(PendingBackend {
 								next_item: Some((backend_key, backend_value)),
 								direction,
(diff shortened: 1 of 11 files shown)


src/btree/mod.rs
```diff
@@ -388,16 +392,16 @@ pub mod commit_overlay {
 			for change in self.changes.iter() {
 				match change {
 					Operation::Set(key, value) => {
-						*bytes += key.len();
-						*bytes += value.len();
+						*bytes += key.value().len();
+						*bytes += value.value().len();
 						overlay.insert(key.clone(), (record_id, Some(value.clone())));
```
hanabi1224, line 397: `key` and `value` are `Arc<Vec<u8>>` now

src/db.rs
```diff
@@ -1313,7 +1364,7 @@ impl IndexedChangeSet {
 			match &change {
 				Operation::Set(k, v) => {
 					*bytes += k.len();
-					*bytes += v.len();
+					*bytes += v.value().len();
 					overlay.indexed.insert(*k, (record_id, Some(v.clone())));
```
hanabi1224, line 1368: `value` is `Arc<Vec<u8>>` now

arkpar:
Thank you for the PR. 
I've ran the stress test with it and this does not seem to affect performance. Memory usage is sligly lower indeed.

src/db.rs
```diff
@@ -56,6 +57,47 @@ const KEEP_LOGS: usize = 16;
 /// Value is just a vector of bytes. Value sizes up to 4Gb are allowed.
 pub type Value = Vec<u8>;
 
+#[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord)]
+pub struct ValuePtr(Arc<Value>);
```
arkpar, line 61: Since this is used for both the key and the value, it should be named in a more generic way.  Something like `SharedVec` or `ArcVec` or `ArcBuf`

Tpt, line 61: + 1 to @arkpar I got confused with the `Ptr`, thinking the wrapped value is maybe `*const Value`.

Also, what about just doing `pub type ValuePtr = Arc<Value>` to get for free everything that is already implented for `Arc<Value>` ? I am not sure the "new type" pattern brings something here.

Tpt (commented):
Thank you for this nice change! I just share the same naming concern as @arkpar

src/db.rs
cheme, line 61: I would also maybe just consider a type alias like `type RcKey = Arc<Key>; type RcValue = Arc<Value>` to reduce the boilerplate implementation.

src/db.rs
```diff
@@ -56,6 +57,47 @@ const KEEP_LOGS: usize = 16;
...
+	}
+}
+
+impl From<Value> for ValuePtr {
+	fn from(value: Value) -> Self {
+		Self(value.into())
+	}
+}
+
+impl<const N: usize> TryFrom<ValuePtr> for [u8; N] {
```
cheme, line 93: is this function used?

src/btree/iter.rs
```diff
@@ -53,8 +57,8 @@ struct PendingBackend {
 pub enum LastKey {
 	Start,
 	End,
-	At(Vec<u8>),
-	Seeked(Vec<u8>),
+	At(ValuePtr),
+	Seeked(ValuePtr),
```
cheme, line 61: I am a bit hesitant to switch this one to rc as it changes the iterator api. Seeked is not really useful, At will be to save one key memory per iterator and a mem copy of the key on all alloc (could use an allocated buffer but not sure if relevant) but the iterator is nowhere near performant or trying to be.
Personally I would rather keep the rc to the change overlay only.

cheme (commented):
Thanks, it sounds like a good idea to switch to Rc in the change overlay.

src/db.rs
hanabi1224: Sure I can rename it.  I use pointer because in the [rust book](https://doc.rust-lang.org/stable/book/ch15-00-smart-pointers.html), `Rc`, `Box` etc are introduced as `smart pointer`

hanabi1224: > what about just doing `pub type ValuePtr = Arc<Value>`

It's used as key type in `HashMap` and `BTreeMap`, but it cannot be queried by `&[u8]` for free

Not working ([playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=2066f2b7f006da3b011d3bc3c8c7860a))
```rust
use std::collections::HashSet;
use std::sync::Arc;

type RcValue = Arc<Vec<u8>>;

fn main() {
    let mut set: HashSet<RcValue> = HashSet::new();
    set.insert(vec![1, 2, 3].into());
    println!("{}", set.contains(&[1, 2, 3][..]));
}
```

Working ([playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=fecb6b574037c56245bba463d7453a9c))
```rust
use std::borrow::Borrow;
use std::collections::HashSet;
use std::sync::Arc;

#[derive(Eq, Hash, PartialEq)]
struct RcValue(Arc<Vec<u8>>);


impl Borrow<[u8]> for RcValue {
	fn borrow(&self) -> &[u8] {
		(&*self.0).borrow()
	}
}

impl From<Vec<u8>> for RcValue {
    fn from(value: Vec<u8>) -> Self {
        Self(value.into())
    }
}

fn main() {
    let mut set: HashSet<RcValue> = HashSet::new();
    set.insert(vec![1, 2, 3].into());
    println!("{}", set.contains(&[1, 2, 3][..]));
}
```

cheme: Makes sense.

src/db.rs
hanabi1224: It's used in tests, let me mark it as `#[cfg(test)]`

src/db.rs
hanabi1224: Renaming done

src/btree/iter.rs
hanabi1224: Could you elaborate on which API to keep unchanged? I just tried reverting it here but it does not change any API signature

cheme: Just iterator `next` and `prev` returns IterResult that used to return plain Vec (as used in the fuzzer, I mean the change to the fuzzer could also be reverted then).

hanabi1224: @cheme Done. Please take another look

hanabi1224:
> Thank you for the PR. I've ran the stress test with it and this does not seem to affect performance. Memory usage is sligtly lower indeed.

Where can I found the stress test code? In my scenario, I have ~70M pairs to ingest upfront so it's a write-only scenario, and I have to commit a batch at 1GB or 2GB to work around the hard coded 16MB write queue size. This change does help my scenery a bit

cheme:
> > Thank you for the PR. I've ran the stress test with it and this does not seem to affect performance. Memory usage is sligtly lower indeed.
> 
> Where can I find the stress test code? In my scenario, I have ~70M pairs to ingest upfront so it's a write-only scenario, and I have to commit batches at 1GB or 2GB to work around the hard-coded 16MB write queue size. This change does help my scenario a bit

cargo run -p parity-db-admin -- stress --help

hanabi1224:
@cheme  thanks!

cheme (approved):
Works for me.

Tpt (approved):

PR #180, opened 2023-02-06
Benchmarks for find_entry

arkpar:
```
$ cargo +nightly bench --features bench bench
    Finished bench [optimized + debuginfo] target(s) in 0.06s
     Running unittests src/lib.rs (target/release/deps/parity_db-21d11d42e7884130)

running 2 tests
test index::test::bench_find_entry     ... bench:          29 ns/iter (+/- 0)
test index::test::bench_find_entry_sse ... bench:          13 ns/iter (+/- 0)
```

Also added a small optimization for `chunk_at`

### DIFF
--- a/src/index.rs
+++ b/src/index.rs
@@ -237,7 +237,6 @@ impl IndexTable {
 		Ok(try_io!(Ok(ptr)))
 	}
 
-	#[inline(always)]
 	#[cfg(target_feature = "sse2")]
 	fn find_entry(
 		&self,
--- a/src/index.rs
+++ b/src/index.rs
@@ -667,8 +667,12 @@ mod test {
 	}
 
 	#[cfg(feature = "bench")]
-	#[bench]
-	fn bench_find_entry(b: &mut Bencher) {
+	fn bench_find_entry_internal<
+		F: Fn(&IndexTable, u64, usize, &[u8; CHUNK_LEN]) -> (Entry, usize),
+	>(
+		b: &mut Bencher,
+		f: F,
+	) {
 		let table =
 			IndexTable { id: TableId(18), map: RwLock::new(None), path: Default::default() };
 		let mut chunk = [0u8; 512];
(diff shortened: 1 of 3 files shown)


src/index.rs
```diff
@@ -287,7 +305,12 @@ impl IndexTable {
...
 
 	#[cfg(any(not(target_feature = "sse2"), test))]
-	fn find_entry_base(&self, key_prefix: u64, sub_index: usize, chunk: &[u8]) -> (Entry, usize) {
+	fn find_entry_base(
+		&self,
+		key_prefix: u64,
+		sub_index: usize,
+		chunk: &[u8; CHUNK_LEN],
+	) -> (Entry, usize) {
 		assert!(chunk.len() >= CHUNK_ENTRIES * 8);
```
cheme, line 314: Is the assert still needed ?

src/index.rs
```diff
@@ -641,4 +670,46 @@ mod test {
...
+			IndexTable::write_entry(&e, i, &mut chunk);
+		}
+
+		let mut index = 0;
+
+		b.iter(|| {
+			let i = index % 64;
+			let x = f(&table, keys[i], 0, &chunk).1;
+			assert_eq!(x, i);
+			index += 1;
```
cheme, line 699: Would make sense to me to have the 64 query under iter (so nb iter don't interfer (actually with this use case it probably does not really matter)).

src/index.rs
```diff
@@ -641,4 +670,46 @@ mod test {
...
+		F: Fn(&IndexTable, u64, usize, &[u8; CHUNK_LEN]) -> (Entry, usize),
+	>(
+		b: &mut Bencher,
+		f: F,
+	) {
+		let table =
+			IndexTable { id: TableId(18), map: RwLock::new(None), path: Default::default() };
+		let mut chunk = [0u8; 512];
+		let mut keys = [0u64; 64];
+		let mut rng = rand::thread_rng();
```
cheme, line 685: Not too sure if we should bench on random content (could be just random from a fix seed).

cheme (approved):

Tpt (approved):
+1 to @cheme suggestions

src/index.rs
arkpar: Under real workload chunks are never full. I'd guess average occupancy is about 33%. With just a few entries base variant will probably be faster, while with 64 entries simd will probably be faster.  This cycle averages it.

src/index.rs
arkpar: OK, even though It does not matter too much what the entry bits are, as long as they are different.

PR #182, opened 2023-02-06
CI: Test aarch64 using Qemu

Tpt:


arkpar:
We are not going to support 32-bit architectures. 32-bit address space is just too small for mmap-based IO. For now the only supported architectures are amd64 (x86_64) and arm64 (aarch64)

Tpt:
I have edited this PR to only test aarch64. It already allowed to find a test compilation bug on aarch64 I introduced with the SIMD work.

arkpar (approved):

cheme (approved):
I discover qemu github action, looks really neat.

PR #183, opened 2023-02-07
Fail compilation if not on amd64 or arm64

Tpt:
Makes clear which platforms are supported

### DIFF
--- a/src/index.rs
+++ b/src/index.rs
@@ -237,7 +237,7 @@ impl IndexTable {
 		Ok(try_io!(Ok(ptr)))
 	}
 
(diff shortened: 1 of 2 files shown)


src/index.rs
```diff
@@ -237,7 +237,7 @@ impl IndexTable {
 		Ok(try_io!(Ok(ptr)))
 	}
 
-	#[cfg(target_feature = "sse2")]
+	#[cfg(target_arch = "x86_64")]
```
Tpt, line 240: x86_64 always have sse2

arkpar (approved):

cheme (approved):

PR #185, opened 2023-02-09
Uniform keys option for stress test.

MattHalpinParity:
Applies the hash function when generating keys and uses a uniform column. It also generates the keys before running the rest of the stress test so the hash function isn't used during the test. I found this useful for profiling.

**Caveats**
Uses memory to store all the keys.
When appending to a database it only generates keys for the current run. Reader threads can read from the whole database so will sometimes end up executing the hash function during the test.

### DIFF
--- a/admin/src/bench/mod.rs
+++ b/admin/src/bench/mod.rs
@@ -167,7 +167,7 @@ impl SizePool {
 		if seed >= self.cache_start {
 			let key_index = seed - self.cache_start;
 			if key_index < self.cached_keys.len() as u64 {
-				return self.cached_keys[key_index as usize];
+				return self.cached_keys[key_index as usize]
 			}
 		}
 
--- a/admin/src/bench/mod.rs
+++ b/admin/src/bench/mod.rs
@@ -137,8 +137,7 @@ impl SizePool {
 	fn cache_keys(&mut self, start: u64, num_keys: u64) {
 		self.cache_start = start;
 		self.cached_keys.clear();
-		for k in 0..num_keys {
-			let index = self.cache_start + k;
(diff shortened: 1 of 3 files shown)


arkpar (approved):

admin/src/bench/mod.rs
```diff
@@ -252,6 +294,19 @@ pub fn run_internal(args: Args, db: Db) {
 		0
 	};
 
+	let mut pool = SizePool::from_histogram(sizes::KUSAMA_STATE_DISTRIBUTION, args.uniform);
+	if args.uniform {
+		println!("Generating uniform keys.");
```
cheme, line 299: I am not too sure when we hit non cached keys, but good to have the code for it.

cheme (approved):

admin/src/bench/mod.rs
MattHalpinParity: Yeah. Reader threads when appending will hit them. Also, writer threads when pruning and appending can hit them sometimes. Might have missed something else.

admin/src/bench/mod.rs
```diff
@@ -105,24 +110,38 @@ impl Stress {
...
 		}
-		SizePool { distribution, total }
+		SizePool { distribution, total, uniform, cache_start: 0, cached_keys: Vec::new() }
+	}
+
+	fn cache_keys(&mut self, start: u64, num_keys: u64) {
+		self.cache_start = start;
+		self.cached_keys.clear();
+		for k in 0..num_keys {
+			let index = self.cache_start + k;
```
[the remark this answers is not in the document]
MattHalpinParity: Done

PR #186, opened 2023-03-02
Siphash key with salt for uniform columns

arkpar:


cheme (approved):
Changes looks good to me.

Not to sure how much < version 4 is obsolete, but probably is.

About choice of sip, the dep is not in the ecosystem yet, but there is already fnv in substrate, but not sure if fnv can be suitable here.

arkpar:
> Not to sure how much < version 4 is obsolete, but probably is.

`LAST_SUPPORTED_VERSION` is already at 4. Earlier versions won't open.

Tpt:
> About choice of sip, the dep is not in the ecosystem yet, but there is already fnv in substrate, but not sure if fnv can be suitable here.

According to [the benchmark](https://cglab.ca/~abeinges/blah/hash-rs/) quoted by the [`rust-fnv`](https://github.com/servo/rust-fnv) project, fnv and sip 2-4 seems to provide similar speed for 32 bytes inputs (our use case). There is no plot about sip 1-3. However FNV does not provide collision attack resistance but I am not sure what is the story about sip 1-3 and if we actually care about it.

arkpar:
siphash is used to avoid collision attacks that my blow up index size. Collision resistance for the siphash itsel however is not important, because the "salt" is random and unknown to the potential attacker.

PR #188, opened 2023-03-07
Preserve logs on write error. 

arkpar:
This fixes a potential DB corruption on write error. When the database is closed with an error, all existing logs must be preserved so that log recovery may fix it on next open.


Tpt (approved):
Thank you! It's sad that the fuzzer has not been able to find this possible error earlier

cheme (approved):

PR #192, opened 2023-03-20
Force value-only iteration in the public API

arkpar:
* Introduced `iter_values` which only iterates value tables sequentially
* Tweaked `check` tool output to include more details for corruptions.

### DIFF
--- a/src/db.rs
+++ b/src/db.rs
@@ -20,7 +20,7 @@
 
 use crate::{
 	btree::{commit_overlay::BTreeChangeSet, BTreeIterator, BTreeTable},
-	column::{hash_key, ColId, Column, IterState, ReindexBatch},
+	column::{hash_key, ColId, Column, IterState, ReindexBatch, ValueIterState},
 	error::{try_io, Error, Result},
 	hash::IdentityBuildHasher,
 	index::PlanOutcome,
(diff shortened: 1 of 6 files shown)


cheme (approved):
Looks good to me, IIRC the value only iterator is not exposed, maybe could be part of check command.

src/table.rs
```diff
@@ -440,6 +442,11 @@ impl ValueTable {
 				return Ok((0, false))
 			}
 
+			if self.multipart && part == 0 && !buf.is_multihead() {
+				// This may only happen during value iteration.
```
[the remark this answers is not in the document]
arkpar: It is not exacly a corruption. Previsouly the table did not care if `multipart` entry is a single entry. The way it is used in column.rs however would not allow that. Perhaps there needs to be a check that disallows inserting small values in multipart tables. I'll leave for another PR though.

src/db.rs
```diff
@@ -20,7 +20,7 @@
 
 use crate::{
 	btree::{commit_overlay::BTreeChangeSet, BTreeIterator, BTreeTable},
-	column::{hash_key, ColId, Column, IterState, ReindexBatch},
+	column::{hash_key, ColId, Column, IterState, ReindexBatch, ValueIterState},
```
Tpt, line 23: Shoudn't we make `ValueIterState` `pub` in `src/lib.rs` to allow its usage from other crates?

Tpt (approved):

src/db.rs
arkpar: Good point, even though public types from private modules are allowed to be leaked. They can be used but can't be named. I've aded it to public exports anyway.

PR #193, opened 2023-03-20
Fix find_entry_sse

arkpar:
In case there are multiple matches for the first group of 4 etries, the `sub_index` check logic would only consider the first match.

### DIFF
--- a/src/index.rs
+++ b/src/index.rs
@@ -274,6 +274,7 @@ impl IndexTable {
 			let target = _mm_set1_epi32(((key_prefix << self.id.index_bits()) >> shift) as i32);
 			let shift_mask = _mm_set_epi64x(0, shift.into());
 			let mut i = (sub_index >> 2) << 2; // We keep an alignment of 4
+			let mut skip = (sub_index - i) as i32;
 			while i + 4 <= CHUNK_ENTRIES {
 				// We load the value 2 by 2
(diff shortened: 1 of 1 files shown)


src/index.rs
```diff
@@ -288,15 +289,13 @@ impl IndexTable {
 				));
 				// We set into current the input low parts
 				let current = _mm_unpacklo_epi64(first_two, last_two);
-				let cmp = _mm_movemask_epi8(_mm_cmpeq_epi32(current, target));
+				let cmp = _mm_movemask_epi8(_mm_cmpeq_epi32(current, target)) >> (skip * 4);
```
cheme, line 292: Is it possible `key_prefix` is 0u64 here and we still fail? (key prefix is 8 first byte from a Key so it is a 64 bit collision, cannot really say if it should be handled considering db salting it sounds like a 2^64 chance guess could be ignored).

cheme (approved):

src/index.rs
cheme: I mean the `>> (skip * 4)` does introduce 0 values and if target is 0 it will match (will probably create a bigger index than what could be expected though.

cheme: Ok I misread this, works fine.

Tpt:
@arkpar Thank you! Sorry for introducing this bug.

PR #195, opened 2023-03-20
Fixes fuzzer: committed but not flushed commits might be written to disk

Tpt:


Tpt:
Closing, this is not normal behavior. I'll investigate why.

PR #196, opened 2023-03-23
Representative query timing

MattHalpinParity:
Added options for controlling writer thread sleeps so that they can pause after a number of commits.
Made reader threads default to not reading pruned values. Old behaviour can be enabled with an option.
These should give a slightly more representative way of measuring query performance.

### DIFF
--- a/admin/src/bench/mod.rs
+++ b/admin/src/bench/mod.rs
@@ -286,31 +286,23 @@ fn reader(
 		if commits == 0 {
 			continue
 		}
-		if args.archive || args.reader_check_pruned {
+		let key = if args.archive || args.reader_check_pruned {
 			let num_keys = commits * COMMIT_SIZE as u64;
-			let key = pool.key(rng.next_u64() % num_keys + seed);
-			match db.get(0, &key).unwrap() {
-				Some(_) => {
-					QUERIES_HIT.fetch_add(1, Ordering::SeqCst);
-				},
-				None => {
-					QUERIES_MISS.fetch_add(1, Ordering::SeqCst);
-				},
-			}
+			pool.key(rng.next_u64() % num_keys + seed)
 		} else {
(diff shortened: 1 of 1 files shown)


admin/src/bench/mod.rs
```diff
@@ -242,26 +262,55 @@ fn writer(
...
+			index += (index / num_commit_values + 1) * COMMIT_PRUNE_SIZE as u64;
+			let key = pool.key(index + seed);
+			match db.get(0, &key).unwrap() {
+				Some(_) => {
+					QUERIES_HIT.fetch_add(1, Ordering::SeqCst);
+				},
+				None => {
+					QUERIES_MISS.fetch_add(1, Ordering::SeqCst);
+				},
+			}
```
arkpar, line 313: To avoid a bit or repetition
```suggestion
		let key = if args.archive || args.reader_check_pruned {
			let num_keys = commits * COMMIT_SIZE as u64;
			pool.key(rng.next_u64() % num_keys + seed)
		} else {
			let num_commit_values = (COMMIT_SIZE - COMMIT_PRUNE_SIZE) as u64;
			let num_keys = commits * num_commit_values;
			let mut index = rng.next_u64() % num_keys;
			index += (index / num_commit_values + 1) * COMMIT_PRUNE_SIZE as u64;
			pool.key(index + seed)
		};
		match db.get(0, &key).unwrap() {
			Some(_) => {
				QUERIES_HIT.fetch_add(1, Ordering::SeqCst);				
			},
			None => {
				QUERIES_MISS.fetch_add(1, Ordering::SeqCst);
			},
		}
```

arkpar (approved):

admin/src/bench/mod.rs
MattHalpinParity: Done

Tpt (approved):

PR #197, opened 2023-03-23
Fuzzer: model more carefully what is written to disk

Tpt:
Allows to more carefully look for data loss

### DIFF
--- a/fuzz/src/lib.rs
+++ b/fuzz/src/lib.rs
@@ -70,8 +70,11 @@ impl<V: Debug> Debug for Layer<V> {
 
 #[derive(Clone, Copy, Debug, Eq, PartialEq)]
 pub enum WrittenState {
+	/// The state has been persisted to disk.
 	Yes,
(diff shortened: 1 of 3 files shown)


arkpar (approved):

fuzz/src/lib.rs
```diff
@@ -48,11 +48,31 @@ pub enum Action<O: Debug> {
...
+			if let Some(v) = v {
+				output.entry(&k, v);
+			}
+		}
+		output.finish()
+	}
+}
+
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
+pub enum WrittenState {
```
cheme, line 72: Even if it is clear from code, a comment on the difference between Yes and Processed would be good from my point of view.

cheme (approved):
LGTM

fuzz/src/lib.rs
Tpt: Done

PR #200, opened 2023-04-16
Improve DB validation tool

arkpar:
* Added an alternative hash column iteration. The index is copied to memory, sorted by disk offset and then iterated. This makes checking large databases 100x faster.
* Also added some code to follow free entry lists and check them for correctness.

### DIFF
--- a/src/column.rs
+++ b/src/column.rs
@@ -931,18 +931,18 @@ impl HashColumn {
 		Ok(())
 	}
 
-	fn dump(&self, log: &Log, check_param: &crate::CheckOptions, col: ColId) -> Result<()> {
-		let start_chunk = check_param.from.unwrap_or(0);
-		let end_chunk = check_param.bound;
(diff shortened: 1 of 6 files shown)


src/column.rs
```diff
@@ -850,29 +857,112 @@ impl HashColumn {
...
+	fn iter_index_fast(
+		&self,
+		log: &Log,
+		mut f: impl FnMut(IterStateOrCorrupted) -> Result<bool>,
+		_start_chunk: u64,
+	) -> Result<()> {
+		let tables = self.tables.read();
+		let index = &tables.index;
+
+		let entries = index.sorted_entries()?;
```
cheme, line 869: Could probably use start_chunk as a restart parameter, but would still need to rebuild sorted entries at all restart so probably something that only make sense to put in place if really needed at some point.

cheme (approved):

Tpt (approved):

PR #202, opened 2023-04-18
Ensure log deletion order

arkpar:
Another take on the problem described in #201


cheme (approved):
Sounds good to me.

cheme:
Is there an interest to add changes to src/file.rs of #201 ?

PR #204, opened 2023-04-20
Limit log cleanup queue

arkpar:
On machines with slow fsync it is possible that the log cleanup queue grows to gigabytes while fsync is in progress. This PR add a limit to the cleanup queue, pausing commits if it grows too much,

Also tweaked error handling for the log recovory. If recovery fails for reasons other than the bad log file we keep the log files around, instead of deleting them and hoping fo the best. If replay is successfull, the logs are deleted right away (after fsync), instead of waiting for the database to be closed.

### DIFF
--- a/src/db.rs
+++ b/src/db.rs
@@ -51,8 +51,13 @@ const MAX_COMMIT_QUEUE_BYTES: usize = 16 * 1024 * 1024;
 const MAX_LOG_QUEUE_BYTES: i64 = 128 * 1024 * 1024;
 // Minimum size of log file before it is considered full.
 const MIN_LOG_SIZE_BYTES: u64 = 64 * 1024 * 1024;
-// Number of log files to keep after flush.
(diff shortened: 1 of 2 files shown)


src/db.rs
```diff
@@ -718,6 +723,13 @@ impl DbInner {
 					}
 					log::debug!(target: "parity-db", "Log queue size: {} bytes", *queue);
 				}
+
+				let max_logs = if self.options.sync_data { MAX_LOG_FILES } else { KEEP_LOGS };
+				let dirty_logs = self.log.num_dirty_logs();
+				if dirty_logs > max_logs {
```
cheme, line 729:
```suggestion
				while self.log.num_dirty_logs() > max_logs {
```
in the sense that cleanup can signal if incomplete clean. But could also be better not to slow down in this case.

Tpt (approved):

cheme (approved):

PR #206, opened 2023-04-24
Fix find_entry for zero partial key

arkpar:
In rare cases when target partial key is all zeroes and the index chunk contains precceding empty entries, these empty entries are returned instead.


MattHalpinParity (approved):
Looks good.

PR #214, opened 2023-07-18
Use mmap IO for value tables

arkpar:
This works much faster when performance is limited by OS page cache. When the database is relatively small and fits entirely in the page cache, making kernel calls becomes too expensive.
Also removed extra copy to a temp buffer when reading values from file. 
Also optimized overlay layout a bit. The top level is a vec now instead of a hashmap.

### DIFF
--- a/src/file.rs
+++ b/src/file.rs
@@ -10,7 +10,12 @@ use crate::{
 };
 use std::sync::atomic::{AtomicU64, Ordering};
 
-const RESERVE_ADDRESS_SPACE: usize = 512 * 1024 * 1024;
+#[cfg(not(test))]
+const RESERVE_ADDRESS_SPACE: usize = 1024 * 1024 * 1024; // 1 Gb
+
+// Use different value for tests to work around docker limits on the test machine.
+#[cfg(test)]
+const RESERVE_ADDRESS_SPACE: usize = 64 * 1024 * 1024; // 64 Mb
 
 #[cfg(target_os = "linux")]
 fn disable_read_ahead(file: &std::fs::File) -> std::io::Result<()> {
--- a/src/file.rs
+++ b/src/file.rs
@@ -10,7 +10,7 @@ use crate::{
 };
 use std::sync::atomic::{AtomicU64, Ordering};
 
-const RESERVE_ADDRESS_SPACE: usize = 1024 * 1024 * 1024;
+const RESERVE_ADDRESS_SPACE: usize = 64 * 1024 * 1024;
 
 #[cfg(target_os = "linux")]
 fn disable_read_ahead(file: &std::fs::File) -> std::io::Result<()> {
--- a/src/file.rs
+++ b/src/file.rs
@@ -147,7 +147,8 @@ impl TableFile {
 				madvise_random(&mut map);
 				*map_and_file = Some((map, file));
 			},
-			Some((_map, file)) => {
+			Some((map, file)) => {
+				std::mem::drop(map);
 				try_io!(file.set_len(capacity * entry_size as u64));
 				let mut m = try_io!(unsafe { memmap2::MmapMut::map_mut(&file) });
 				madvise_random(&mut m);
--- a/src/file.rs
+++ b/src/file.rs
@@ -39,11 +39,7 @@ fn disable_read_ahead(_file: &std::fs::File) -> std::io::Result<()> {
 #[cfg(unix)]
 pub fn madvise_random(map: &mut memmap2::MmapMut) {
 	unsafe {
-		libc::madvise(
-			map.as_mut_ptr() as _,
-			map.len(),
-			libc::MADV_RANDOM,
-		);
+		libc::madvise(map.as_mut_ptr() as _, map.len(), libc::MADV_RANDOM);
 	}
 }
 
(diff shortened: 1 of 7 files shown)


src/file.rs
```diff
@@ -193,26 +138,36 @@ impl TableFile {
...
+				madvise_random(&mut m);
+				*map = m;
+			},
 		}
-		try_io!(file.as_ref().unwrap().set_len(capacity * entry_size as u64));
 		Ok(())
 	}
 
 	pub fn flush(&self) -> Result<()> {
-		if let Some(file) = self.file.read().as_ref() {
```
cheme, line 207: in https://github.com/cheme/parity-db/blob/3b49e8724f30573570faab9f8902ad7b65211bc1/src/file.rs#L470 I did use flush range to avoid flushing the start of the file (probably the header), but tbh it does not really make sense to me right now, just mentioning in case there may be some use to it.

src/file.rs
```diff
@@ -193,26 +138,36 @@ impl TableFile {
...
+			None => {
+				let file = self.create_file()?;
+				try_io!(file.set_len(capacity * entry_size as u64));
+				let mut map = try_io!(unsafe { memmap2::MmapMut::map_mut(&file) });
+				madvise_random(&mut map);
+				*map_and_file = Some((map, file));
+			},
+			Some((map, file)) => {
+				try_io!(file.set_len(capacity * entry_size as u64));
+				let mut m = try_io!(unsafe { memmap2::MmapMut::map_mut(&*file) });
```
cheme, line 152: Do we need to drop the existing mmap before creating this one?
(looks good if it works fine this way)

cheme (commented):
Lgtm,
I remember having a branch with a trait to be able to choose file or mmap depending on use case as in https://github.com/cheme/parity-db/blob/file_mmap2/src/file.rs.
There was some perf issue, but I was locking write on the mmap, maybe locking read as here, thing would have been good.
I remember bigger issue was grow, not exactly sure why but I ended up with https://github.com/cheme/parity-db/blob/3b49e8724f30573570faab9f8902ad7b65211bc1/src/file.rs#L420 at the time (madvise is missing and mflush is useless).

src/file.rs
arkpar: That's probably a good idea

src/file.rs
arkpar: Header for a value table is enacted as a regular entry, so it should be synced as well.

arkpar:
Regarding `grow`, it does take some time indeed to create a new mapping. `mremap` only works on linux. One solution is to simply reserve a lot of address space for each file with the initial mapping. Then, as file grows, this additional address space can be used without remapping. I've left this optimization for later though. Current version with creating a new map each time is quick enough.

src/file.rs
```diff
@@ -36,42 +38,34 @@ fn disable_read_ahead(_file: &std::fs::File) -> std::io::Result<()> {
...
 pub struct TableFile {
-	pub file: RwLock<Option<std::fs::File>>,
+	pub map: RwLock<Option<(memmap2::MmapMut, std::fs::File)>>,
 	pub path: std::path::PathBuf,
 	pub capacity: AtomicU64,
 	pub id: TableId,
 }
 
+fn map_len(file_len: u64) -> usize {
+	file_len as usize + RESERVE_ADDRESS_SPACE
```
cheme, line 62: Wonder if we also should try to align the size?

src/log.rs
```diff
@@ -447,6 +549,10 @@ pub struct IndexLogOverlay {
 // We use identity hash for value overlay/log records so that writes to value tables are in order.
 #[derive(Debug, Default)]
 pub struct ValueLogOverlay {
+	pub map: HashMap<u64, (u64, Vec<u8>)>, // index -> (record_id, entry)
```
cheme, line 552: Not too sure why we remove the BuildIdHash here

cheme (commented):
aarch64 test did fail, not sure why, I did not try to relaunch it (looks like it timeout, hope it is not an atomic aarch64 only hard to debug issue).

src/log.rs
arkpar: The only reason for it is so that log changes are ordered by entry index when written to the disk. So it only makes sense to use in the local overlay. Ideally, local overlay should maintain a an additional sorted list for this. Identity hash for non-random indicies degrades in-memory performance.

src/file.rs
arkpar: I'm made is so that file size is always aligned.

src/log.rs
cheme: make sense, thanks for the explanation.

src/file.rs
```diff
@@ -4,12 +4,19 @@
...
 	table::TableId,
 };
 use std::sync::atomic::{AtomicU64, Ordering};
 
+#[cfg(not(test))]
+const RESERVE_ADDRESS_SPACE: usize = 1024 * 1024 * 1024; // 1 Gb
+
+// Use different value for tests to work around docker limits on the test machine.
+#[cfg(test)]
+const RESERVE_ADDRESS_SPACE: usize = 64 * 1024 * 1024; // 64 Mb
```
cheme, line 18: Wondering if this could be configurable (probably these kind of limitation might exist on some cloud infra).
Maybe even a bit more granular (eg more reserved for log).

cheme (approved):

src/file.rs
arkpar: We'll give it a round of testing on the parity infrastructure. I've tested on my arm64 vm and had no issues.

src/file.rs
```diff
@@ -108,111 +110,87 @@ impl TableFile {
...
+				try_io!(file.set_len(new_len));
+				if map.len() < new_len as usize {
+					let mut new_map = try_io!(unsafe {
+						memmap2::MmapOptions::new().len(map_len(new_len)).map_mut(&*file)
+					});
+					madvise_random(&mut new_map);
+					let old_map = std::mem::replace(map, new_map);
+					try_io!(old_map.flush());
+					// Leak the old mapping as there might be concurrent readers.
+					std::mem::forget(old_map);
```
nazar-pc, line 173: Why is this leak not an issue? I'm trying to identify the root cause of https://github.com/paritytech/parity-db/issues/233 and this looks like a candidate.

PR #216, opened 2023-07-21
Remove log overlay

arkpar:
This introduce a major chage in the WAL format. Current design writes WAL in the format that matches the on-disk format. I.e. index pages and value table entries. This seemed like a good idea at the time, as it allowed to split the work more efficiently between threads, and incorporate the reindex process. However it turned out that this approach has some drawbacks:

1. Log records are bigger in size. Write amplification is high.
2. Possibility of page evicton. There's significant time between planning and enactment phases. Planning phase would load a page from disk into OS page cache (to e.g. check if the key matches), and write an update to the log. By the time that log is eanced the OS might have already evicted the page and would need to load it again.

This new design changes the log records to contain high level DB operations, rather than the low level table updates. Operations are now executed at the enactment phase. `CommitOverlay` holds newly inserted key-values until they are enacted.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "parity-db"
-version = "0.4.10"
+version = "0.4.11"
 authors = ["Parity Technologies <admin@parity.io>"]
 edition = "2021"
 license = "MIT OR Apache-2.0"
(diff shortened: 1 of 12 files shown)


arkpar:
After some consideraton, this is not really sound w.r.t. reference handling. Operations such ash reference increment and decrement must include the current reference value in the log, as they can't be enacted twice after an unclean shutdown and log replay. So this needs more work. It might be worth it to have a separate table for keeping ref counts. Only entries with more than one RC would go there, so the table would be rather small.

This PR needs more work, so I'm closing it for now.

PR #221, opened 2023-09-04
Make `madvise_random` compatible with non-Unix OS

### LINKED ISSUE #220: `madvise_random` function in v0.4.10 has different signatures for Unix and non-Unix systems
The `madvise_random` function for both Unix and non-Unix systems had the same signature in v0.4.9:

https://github.com/paritytech/parity-db/blob/v0.4.9/src/index.rs#L145-L157

Since the v0.4.10 function signatures are different:

https://github.com/paritytech/parity-db/blob/v0.4.10/src/file.rs#L47-L54

This yields build error when building for, e.g., Windows.

## Possible solution

Make non-Unix signature the same as the Unix one.

```rust
#[cfg(not(unix))]
pub fn madvise_random(_map: &mut memmap2::MmapMut) {}
```

shamilsan:
Fixes #220.

### DIFF
--- a/src/file.rs
+++ b/src/file.rs
@@ -51,7 +51,7 @@ pub fn madvise_random(map: &mut memmap2::MmapMut) {
 }
 
 #[cfg(not(unix))]
-pub fn madvise_random(_id: TableId, _map: &mut memmap2::MmapMut) {}
+pub fn madvise_random(_map: &mut memmap2::MmapMut) {}
 
 const GROW_SIZE_BYTES: u64 = 256 * 1024;
 


arkpar (approved):

nazar-pc:
@arkpar could you release this as a new version? Just hit this in Windows CI. Also would be great to add macOS and Windows to CI here.

arkpar:
published as v0.4.11

nazar-pc:
How I'm curious why not using https://docs.rs/memmap2/latest/memmap2/struct.Advice.html with `.advise()` methods on with memmap2? It does the same thing internally, but spared you from using unsafe `libc::madvise` (and takes care of alignment requirement for its argument).

arkpar:
That function was not available in `memmap2-rs` back when this code was originally written. It makes sense to switch now indeed.

PR #222, opened 2023-09-13
CI for windows and macos

arkpar:
Disable mmap address space reservation on windows as it is not supported by memmap2-rs

### DIFF
--- a/src/file.rs
+++ b/src/file.rs
@@ -93,7 +93,6 @@ impl TableFile {
 				capacity = len / entry_size as u64;
 			}
 			let mut map = mmap(&file, len as usize)?;
-			madvise_random(&mut map);
 			Some((map, file))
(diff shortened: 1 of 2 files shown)


src/file.rs
```diff
@@ -84,8 +92,7 @@ impl TableFile {
 			} else {
 				capacity = len / entry_size as u64;
 			}
-			let mut map =
-				try_io!(unsafe { memmap2::MmapOptions::new().len(map_len(len)).map_mut(&file) });
+			let mut map = mmap(&file, len as usize)?;
 			madvise_random(&mut map);
```
MattHalpinParity, line 96: Looks good. Don't think this line is needed anymore.

MattHalpinParity (approved):

PR #225, opened 2023-10-06
Force alignment for all chunk buffers

### LINKED ISSUE #223: Misaligned pointer dereference in `chunk_entries_at` and `transmute_chunk`
### The source of unsoundness
Hi, we consider that the following two functions could have misaligned pointer dereference and lead to UB:
https://github.com/paritytech/parity-db/blob/4ac2aca38c6984f8a48080875590a687041a39d8/src/index.rs#L255-L258
At line 258 of `chunk_entries_at`, 
https://github.com/paritytech/parity-db/blob/4ac2aca38c6984f8a48080875590a687041a39d8/src/index.rs#L418-L422
and at line 419 of `transmute_chunk`,
they both tried to convert the type aligned to 1 byte to the type aligned to 8 bytes. Please check and would be happy to have any discussion:)

arkpar:
Closes #223

### DIFF
--- a/src/index.rs
+++ b/src/index.rs
@@ -287,7 +287,6 @@ impl IndexTable {
 			return self.find_entry_base(key_prefix, sub_index, chunk)
 		}
 		unsafe {
-			let chunk = &chunk.0;
 			let target = _mm_set1_epi32(pk as i32);
 			let shift_mask = _mm_set_epi64x(0, shift.into());
 			let mut i = (sub_index >> 2) << 2; // We keep an alignment of 4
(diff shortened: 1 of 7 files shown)


src/index.rs
```diff
@@ -27,9 +27,12 @@ const META_SIZE: usize = 16 * 1024; // Contains header and column stats
 const ENTRY_BITS: u8 = 64;
 pub const ENTRY_BYTES: usize = ENTRY_BITS as usize / 8;
 
-const EMPTY_CHUNK: Chunk = [0u8; CHUNK_LEN];
+const EMPTY_CHUNK: Chunk = Chunk([0u8; CHUNK_LEN]);
+const EMPTY_ENTRIES: [Entry; CHUNK_ENTRIES] = [Entry::empty(); CHUNK_ENTRIES];
 
-pub type Chunk = [u8; CHUNK_LEN];
+#[repr(C, align(8))]
+#[derive(PartialEq, Eq, Clone, Copy, Debug)]
```
cheme, line 34: Is the impact of not being `Copy` big ? (but not sure if we want to force explicit clone here).

src/lib.rs
```diff
@@ -35,3 +35,6 @@ pub type Key = [u8; KEY_SIZE];
 
 #[cfg(not(any(target_arch = "x86_64", target_arch = "aarch64")))]
 compile_error!("parity-db only supports x86_64 and aarch64");
+
+#[cfg(not(target_endian = "little"))]
+compile_error!("parity-db only supports little-endian platforms");
```
cheme, line 40: Make sense (keeping support would require a CI test indeed).

cheme (approved):

src/lib.rs
arkpar: Yeah, it is not tested properly. The platform restriction of x86_64 and aarch64 limits it to litle endian anyway.

src/index.rs
arkpar: Make it explicit. Also changed `transmute_chunk` to operate on references, as it allows to eliminate a copy in some places.

MattHalpinParity (approved):
LGTM

PR #230, opened 2023-10-16
Fix value table iteration vs log processing race

### LINKED ISSUE #227: Panic on TableFile::slice_at.
Steps to reproduce: 
1. Reinstate the selected lines of [this test](https://github.com/ChainSafe/forest/pull/3504/commits/66687bdad028d526dde063aca5a3adcedfa13a78#diff-5475d3517d34591f55cbd3c91700c3a8f7051f1d076bd45bbfa13143bfdd0893R446-R448)
2. Run the test several times, this reproduces quite consistently on my machine. 

Version: 0.4.11.

arkpar:
Fixes #227

### DIFF
--- a/fuzz/src/lib.rs
+++ b/fuzz/src/lib.rs
@@ -46,6 +46,7 @@ pub enum Action<O: Debug> {
 	Restart,
 	IterPrev,
 	IterNext,
+	IterValues,
 }
 
 #[derive(Clone)]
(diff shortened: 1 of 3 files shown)


cheme (approved):
change filled to written makes sense to me.
iiuc the _iteration_lock is needed for the linked issue where  get_keys is doing iteration on key concurrently to db normal usage, and change in test is needed to reflect filled.
Maybe would be good to have such use case in testing or fuzzing (iter_while calls against standard usage of db), but not sure what would be appropriate here.

Issue #233, opened 2024-01-04
Memory is "leaking" on Windows and leads to crashes


nazar-pc:
After many complaints from users of our Substrate-based chain I did some experiments and discovered that with ParityDb application uses a lot of memory on Windows until it runs out of memory, while RocksDb works fine.

I tried both mimalloc allocator and default system allocator with the same result.

Errors typically look something like this:
```
2024-01-04T15:49:21.505650Z  WARN parity-db: Background worker error: IO Error: The paging file is too small for this operation to complete. (os error 1455)    
```

Interestingly enough, task manager shows very reasonable memory usage of the app in the process list, while there is a huge total memory usage in Performance tab.

In Sysinternals Process Explorer I see that while the app uses ~3G of "Working Set" memory (similarly to what task manager shows), "Private Bytes" is over 10x more (again, not happening with RocksDb).

Not sure how to narrow it down more yet, so open for suggestions and tests, but this is a pretty big issue that I'd like to help resolve.

nazar-pc:
Testing it now, will take some time to get results.

I also noticed that read-ahead is not disabled on Windows, even though it is possible: https://github.com/nazar-pc/parity-db/commit/388400e9d088dbf8f21b4b08a7dbfd1a5ef9fd06
It didn't make a difference for memory usage, but maybe still helpful, should I send a PR?

arkpar:
Sure. It should help with performance.

arkpar:
Published as 0.4.13

PR #234, opened 2024-01-05
Lock memory map when reading from file

arkpar:


cheme (approved):
Make sense.
I remember having issues testing the grow method of windows on a linux and my code at the time was droping before creating but I think it solve nothing in this case the lock looks needed.
Just wondering if the lock is stricly linux when using linux grow method (overall seems safer to have it)?

nazar-pc:
I have a high degree of confidence this fixes https://github.com/paritytech/parity-db/issues/233. Previously in less than an hour it was crashing, now I don't see any indication of the problem happening at all.

Would be awesome to get this in a patch version.

PR #237, opened 2024-04-04
file size limit?

cheme:
Been ask a few time for a size limit on parity db files.

In itself it does not really make sense, but practically I believe that many tools are pretty bad at using big files (like not even chunking them).
Still this should not be default, but can be of use.

So I try to do a quick implementation to see if it can be a good idea to have.

Got this branch that seems to run the stress test: option max_file_size is in megabyte.

A few todo to handle and test to modify (so we have some runing with multiple files) and this could be good to go (an maybe use smallvec instead of vec).

cc\ @arkpar @MattHalpinParity @PierreBesson

### DIFF
--- a/src/file.rs
+++ b/src/file.rs
@@ -14,6 +14,15 @@ trait OpenOptionsExt {
 	fn disable_read_ahead(&mut self) -> &mut Self;
 }
 
+#[inline]
+fn offset_to_file_index(offset: u64, max_size: Option<usize>) -> (u64, usize) {
+	if let Some(m) = max_size {
+		(offset % m as u64, offset as usize / m)
+	} else {
+		(offset, 0)
+	}
+}
+
 impl OpenOptionsExt for std::fs::OpenOptions {
 	#[cfg(not(windows))]
 	fn disable_read_ahead(&mut self) -> &mut Self {
@@ -66,13 +75,14 @@ pub fn madvise_random(_map: &mut memmap2::MmapMut) {}
 
 #[cfg(not(windows))]
 fn mmap(file: &std::fs::File, len: usize) -> Result<memmap2::MmapMut> {
+	// TODO pass len not 0 and limit reserve to max file size
 	#[cfg(not(test))]
 	const RESERVE_ADDRESS_SPACE: usize = 1024 * 1024 * 1024; // 1 Gb
 														 // Use a different value for tests to work around docker limits on the test machine.
 	#[cfg(test)]
 	const RESERVE_ADDRESS_SPACE: usize = 64 * 1024 * 1024; // 64 Mb
 
-	let map_len = len + RESERVE_ADDRESS_SPACE;
+	let map_len = len + RESERVE_ADDRESS_SPACE; // TODO should be max??
 	let mut map = try_io!(unsafe { memmap2::MmapOptions::new().len(map_len).map_mut(file) });
 	madvise_random(&mut map);
 	Ok(map)
@@ -87,69 +97,94 @@ const GROW_SIZE_BYTES: u64 = 256 * 1024;
 
 #[derive(Debug)]
 pub struct TableFile {
-	pub map: RwLock<Option<(memmap2::MmapMut, std::fs::File)>>,
-	pub path: std::path::PathBuf,
+	pub map: RwLock<Vec<(memmap2::MmapMut, std::fs::File)>>,
+	pub path_base: std::path::PathBuf,
 	pub capacity: AtomicU64,
 	pub id: TableId,
+	pub max_size: Option<usize>,
 }
 
 impl TableFile {
-	pub fn open(filepath: std::path::PathBuf, entry_size: u16, id: TableId) -> Result<Self> {
+	pub fn open(
+		path_base: std::path::PathBuf,
+		entry_size: u16,
+		id: TableId,
+		max_file_size: Option<usize>,
+	) -> Result<Self> {
 		let mut capacity = 0u64;
-		let map = if std::fs::metadata(&filepath).is_ok() {
-			let file = try_io!(std::fs::OpenOptions::new()
-				.read(true)
-				.write(true)
-				.disable_read_ahead()
-				.open(filepath.as_path()));
-			try_io!(disable_read_ahead(&file));
-			let len = try_io!(file.metadata()).len();
-			if len == 0 {
-				// Preallocate.
-				capacity += GROW_SIZE_BYTES / entry_size as u64;
-				try_io!(file.set_len(GROW_SIZE_BYTES));
+		let mut maps = Vec::new();
+		for i in 0.. {
+			let mut filepath = path_base.clone();
+			filepath.push(id.file_name(max_file_size.is_some().then(|| i)));
+			if std::fs::metadata(&filepath).is_ok() {
+				let file = try_io!(std::fs::OpenOptions::new()
+					.read(true)
+					.write(true)
+					.disable_read_ahead()
+					.open(filepath.as_path()));
+				try_io!(disable_read_ahead(&file));
+				let len = try_io!(file.metadata()).len();
+				if len == 0 {
+					// Preallocate.
+					capacity += GROW_SIZE_BYTES / entry_size as u64;
+					try_io!(file.set_len(GROW_SIZE_BYTES));
+				} else {
+					capacity = len / entry_size as u64;
+				}
+				let map = mmap(&file, len as usize)?;
+				maps.push((map, file));
 			} else {
-				capacity = len / entry_size as u64;
+				break;
+			};
+			if max_file_size.is_none() {
+				break;
 			}
-			let map = mmap(&file, len as usize)?;
-			Some((map, file))
+		}
+		let max_size = if let Some(m) = max_file_size {
+			let m = m * 1024 * 1024;
+			Some((m / entry_size as usize) * entry_size as usize)
 		} else {
 			None
 		};
 		Ok(TableFile {
-			path: filepath,
-			map: RwLock::new(map),
+			path_base,
+			map: RwLock::new(maps),
 			capacity: AtomicU64::new(capacity),
 			id,
+			max_size,
 		})
 	}
 
-	fn create_file(&self) -> Result<std::fs::File> {
+	fn create_file(&self, index: Option<u32>) -> Result<std::fs::File> {
 		log::debug!(target: "parity-db", "Created value table {}", self.id);
+		let mut path = self.path_base.clone();
+		path.push(self.id.file_name(index));
 		let file = try_io!(std::fs::OpenOptions::new()
 			.create(true)
 			.read(true)
 			.write(true)
 			.disable_read_ahead()
-			.open(self.path.as_path()));
+			.open(&path));
 		try_io!(disable_read_ahead(&file));
 		Ok(file)
 	}
 
 	pub fn read_at(&self, buf: &mut [u8], offset: u64) -> Result<()> {
+		let (offset, file_index) = offset_to_file_index(offset, self.max_size);
 		let offset = offset as usize;
 		let map = self.map.read();
-		let (map, _) = map.as_ref().unwrap();
+		let (map, _) = map.get(file_index).unwrap();
 		buf.copy_from_slice(&map[offset..offset + buf.len()]);
 		Ok(())
 	}
 
 	#[cfg(not(feature = "loom"))]
 	pub fn slice_at(&self, offset: u64, len: usize) -> MappedBytesGuard {
+		let (offset, file_index) = offset_to_file_index(offset, self.max_size);
 		let offset = offset as usize;
 		let map = self.map.read();
 		parking_lot::RwLockReadGuard::map(map, |map| {
-			let (map, _) = map.as_ref().unwrap();
+			let (map, _) = map.get(file_index).unwrap();
 			&map[offset..offset + len]
 		})
 	}
@@ -164,7 +199,8 @@ impl TableFile {
 
 	pub fn write_at(&self, buf: &[u8], offset: u64) -> Result<()> {
 		let map = self.map.read();
-		let (map, _) = map.as_ref().unwrap();
+		let (offset, file_index) = offset_to_file_index(offset, self.max_size);
+		let (map, _) = map.get(file_index).unwrap();
 		let offset = offset as usize;
 
 		// Nasty mutable pointer cast. We do ensure that all chunks that are being written are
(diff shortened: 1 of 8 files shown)


BulatSaif:
Is this change backward compatible? For example, if I create a snapshot for a node with the `max_file_size` parameter, and someone downloads it and starts the node without specifying `max_file_size`.

cheme:
> Is this change backward compatible? For example, if I create a snapshot for a node with the `max_file_size` parameter, and someone downloads it and starts the node without specifying `max_file_size`.

the parameter cannot be change, it is written in metadata, so should error when trying to open with a wrong parameter.
so depends on how it will be integrated with substrate, but there is no switching this max size.

arkpar:
Great stuff.
Ideally we could map files to a sequeantial address space, so that looking up mapping addres based on chunk/entry won't be required. But the benefits probably will be negligible.

One thing I'm concerned about is hitting the OS limit for memory maps (`/proc/sys/vm/max_map_count`). We should run polkadot archive DB with this and see how many mmaps it creates and how close we are to the limit.

src/file.rs
```diff
@@ -66,13 +75,14 @@ pub fn madvise_random(_map: &mut memmap2::MmapMut) {}
...
 fn mmap(file: &std::fs::File, len: usize) -> Result<memmap2::MmapMut> {
+	// TODO pass len not 0 and limit reserve to max file size
 	#[cfg(not(test))]
 	const RESERVE_ADDRESS_SPACE: usize = 1024 * 1024 * 1024; // 1 Gb
 														 // Use a different value for tests to work around docker limits on the test machine.
 	#[cfg(test)]
 	const RESERVE_ADDRESS_SPACE: usize = 64 * 1024 * 1024; // 64 Mb
 
-	let map_len = len + RESERVE_ADDRESS_SPACE;
+	let map_len = len + RESERVE_ADDRESS_SPACE; // TODO should be max??
```
arkpar, line 85: What do you mean with this TODO here?

src/file.rs
```diff
@@ -180,44 +217,72 @@ impl TableFile {
 
 	pub fn grow(&self, entry_size: u16) -> Result<()> {
 		let mut map_and_file = self.map.write();
-		let new_len = match map_and_file.as_mut() {
-			None => {
-				let file = self.create_file()?;
-				let len = GROW_SIZE_BYTES;
-				try_io!(file.set_len(len));
-				let map = mmap(&file, 0)?;
```
cheme, line 188: Here I switched from `mmap(&file, 0)` to `mmap(&file, len)`.

cheme:
> One thing I'm concerned about is hitting the OS limit for memory maps (/proc/sys/vm/max_map_count). We should run polkadot archive DB with this and see how many mmaps it creates and how close we are to the limit.

yes, sure is convenient being able to define 1mb files (for testing), but I would not really expect something lower than 512Mb as max size, still with a db of 200gb, that would be more than 400 open mmap: clearly something to test.

cheme:
this patch can be use to run it on polkadot-sdk:
```
diff --git a/Cargo.lock b/Cargo.lock
index c3cf0b953f..5b9c5a0226 100644
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -11438,7 +11438,7 @@ checksum = "4e69bf016dc406eff7d53a7d3f7cf1c2e72c82b9088aac1118591e36dd2cd3e9"
 dependencies = [
  "bitcoin_hashes 0.13.0",
  "rand",
- "rand_core 0.6.4",
+ "rand_core 0.5.1",
  "serde",
  "unicode-normalization",
 ]
@@ -11451,9 +11451,7 @@ checksum = "16b56e3a2420138bdb970f84dfb9c774aea80fa0e7371549eedec0d80c209c67"
 
 [[package]]
 name = "parity-db"
-version = "0.4.12"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "59e9ab494af9e6e813c72170f0d3c1de1500990d62c97cc05cc7576f91aa402f"
+version = "0.4.13"
 dependencies = [
  "blake2 0.10.6",
  "crc32fast",
@@ -11467,6 +11465,7 @@ dependencies = [
  "rand",
  "siphasher",
  "snap",
+ "winapi",
 ]
 
 [[package]]
diff --git a/Cargo.toml b/Cargo.toml
index 067b65ff22..69e4ecdbdf 100644
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -634,3 +634,6 @@ wasmi = { opt-level = 3 }
 x25519-dalek = { opt-level = 3 }
 yamux = { opt-level = 3 }
 zeroize = { opt-level = 3 }
+
+[patch.crates-io]
+parity-db = { git = "https://github.com/cheme/parity-db.git", branch = "split_files" }
diff --git a/polkadot/node/service/Cargo.toml b/polkadot/node/service/Cargo.toml
index 9688ab5564..b85ff8a023 100644
--- a/polkadot/node/service/Cargo.toml
+++ b/polkadot/node/service/Cargo.toml
@@ -89,7 +89,7 @@ serde_json = { workspace = true, default-features = true }
 thiserror = { workspace = true }
 kvdb = "0.13.0"
 kvdb-rocksdb = { version = "0.19.0", optional = true }
-parity-db = { version = "0.4.12", optional = true }
+parity-db = { version = "0.4.13", optional = true }
 codec = { package = "parity-scale-codec", version = "3.6.1" }
 parking_lot = "0.12.1"
 bitvec = { version = "1.0.1", optional = true }
diff --git a/polkadot/node/subsystem-util/Cargo.toml b/polkadot/node/subsystem-util/Cargo.toml
index 79a6a75e4c..db35ddf34f 100644
--- a/polkadot/node/subsystem-util/Cargo.toml
+++ b/polkadot/node/subsystem-util/Cargo.toml
@@ -40,7 +40,7 @@ sp-keystore = { path = "../../../substrate/primitives/keystore" }
 sc-client-api = { path = "../../../substrate/client/api" }
 
 kvdb = "0.13.0"
-parity-db = { version = "0.4.12" }
+parity-db = { version = "0.4.13" }
 
 [dev-dependencies]
 assert_matches = "1.4.0"
diff --git a/substrate/bin/node/bench/Cargo.toml b/substrate/bin/node/bench/Cargo.toml
index 49485fe2a1..422f0fc1c0 100644
--- a/substrate/bin/node/bench/Cargo.toml
+++ b/substrate/bin/node/bench/Cargo.toml
@@ -41,7 +41,7 @@ tempfile = "3.1.0"
 fs_extra = "1"
 rand = { version = "0.8.5", features = ["small_rng"] }
 lazy_static = "1.4.0"
-parity-db = "0.4.12"
+parity-db = "0.4.13"
 sc-transaction-pool = { path = "../../../client/transaction-pool" }
 sc-transaction-pool-api = { path = "../../../client/transaction-pool/api" }
 futures = { version = "0.3.30", features = ["thread-pool"] }
diff --git a/substrate/client/db/Cargo.toml b/substrate/client/db/Cargo.toml
index 57ee1a8ad3..fd24c73b0a 100644
--- a/substrate/client/db/Cargo.toml
+++ b/substrate/client/db/Cargo.toml
@@ -25,7 +25,7 @@ kvdb-memorydb = "0.13.0"
 kvdb-rocksdb = { version = "0.19.0", optional = true }
 linked-hash-map = "0.5.4"
 log = { workspace = true, default-features = true }
-parity-db = "0.4.12"
+parity-db = "0.4.13"
 parking_lot = "0.12.1"
 sc-client-api = { path = "../api" }
 sc-state-db = { path = "../state-db" }
diff --git a/substrate/client/db/src/parity_db.rs b/substrate/client/db/src/parity_db.rs
index b7068f2430..accf768908 100644
--- a/substrate/client/db/src/parity_db.rs
+++ b/substrate/client/db/src/parity_db.rs
@@ -41,6 +41,7 @@ pub fn open<H: Clone + AsRef<[u8]>>(
 	upgrade: bool,
 ) -> parity_db::Result<std::sync::Arc<dyn Database<H>>> {
 	let mut config = parity_db::Options::with_columns(path, NUM_COLUMNS as u8);
+	config.max_file_size = Some(128);
 
 	match db_type {
 		DatabaseType::Full => {
diff --git a/substrate/client/statement-store/Cargo.toml b/substrate/client/statement-store/Cargo.toml
index 8ca6d11dbe..6ded47f238 100644
--- a/substrate/client/statement-store/Cargo.toml
+++ b/substrate/client/statement-store/Cargo.toml
@@ -18,7 +18,7 @@ targets = ["x86_64-unknown-linux-gnu"]
 [dependencies]
 log = { workspace = true, default-features = true }
 parking_lot = "0.12.1"
-parity-db = "0.4.12"
+parity-db = "0.4.13"
 tokio = { version = "1.22.0", features = ["time"] }
 sp-statement-store = { path = "../../primitives/statement-store" }
 prometheus-endpoint = { package = "substrate-prometheus-endpoint", path = "../../utils/prometheus" }
 ```

File: src/index.rs [hunk 3]
```diff
@@ -208,31 +229,59 @@ impl std::fmt::Display for TableId {
...
+				Err(e) => return Err(Error::Io(e)),
+				Ok(file) => file,
+			};
 
-		try_io!(file.set_len(file_size(id.index_bits())));
-		let mut map = try_io!(unsafe { memmap2::MmapMut::map_mut(&file) });
-		madvise_random(&mut map);
-		log::debug!(target: "parity-db", "Opened existing index {}", id);
-		Ok(Some(IndexTable { id, path, map: RwLock::new(Some(map)) }))
+			try_io!(file.set_len(file_size(id.index_bits())));
```
MattHalpinParity, line 255: Looks like all the index files are getting the original total size rather than smaller size based on max_chunks.

src/index.rs
```diff
@@ -526,23 +585,29 @@ impl IndexTable {
...
+			for i in map_len..file_index + 1 {
+				let mut path = self.path_base.clone();
+				path.push(self.id.file_name(self.max_size.is_some().then(|| i as u32)));
+				let file = try_io!(std::fs::OpenOptions::new()
+					.write(true)
+					.read(true)
+					.create_new(true)
+					.open(path.as_path()));
+				log::debug!(target: "parity-db", "Created new index {}", self.id);
+				try_io!(file.set_len(file_size(self.id.index_bits())));
```
MattHalpinParity, line 601: Same file size issue as above.

src/ref_count.rs
```diff
@@ -142,31 +163,60 @@ impl RefCountTable {
...
+					if i == 0 {
+						return Ok(None)
+					} else {
+						break
+					},
+				Err(e) => return Err(Error::Io(e)),
+				Ok(file) => file,
+			};
+
+			try_io!(file.set_len(file_size(id.index_bits())));
```
MattHalpinParity, line 186: Same as with index.

src/file.rs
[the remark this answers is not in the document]
arkpar: `RESERVE_ADDRESS_SPACE` is supposed to be additional reserved space. I.e. if `RESERVE_ADDRESS_SPACE` is 1024Mb and we try to map 1025Mb it should reserve additional space up to 2048Mb so that when next time we try to map 1026Mb it could use the same space.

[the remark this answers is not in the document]
arkpar: if `max_len` is set, it should be used as max map size instead of `RESERVE_ADDRESS_SPACE`

src/file.rs
```diff
@@ -87,84 +97,111 @@ const GROW_SIZE_BYTES: u64 = 256 * 1024;
 
 #[derive(Debug)]
 pub struct TableFile {
-	pub map: RwLock<Option<(memmap2::MmapMut, std::fs::File)>>,
-	pub path: std::path::PathBuf,
+	pub map: RwLock<Vec<(memmap2::MmapMut, std::fs::File)>>,
```
arkpar, line 100: should be called `maps` or `mem_maps`

src/file.rs
```diff
@@ -180,44 +217,72 @@ impl TableFile {
...
+				if self.max_size == Some(len as usize) {
+					(0, true)
+				} else {
+					(len, false)
 				}
-				new_len
 			},
 		};
-		let capacity = new_len / entry_size as u64;
+		let per_page_capacity = self.max_size.map(|m| m / entry_size as usize).unwrap_or(0);
```
arkpar, line 232: `entries_per_file`. Let's not introduce terms such as "page" here.

src/file.rs
```diff
@@ -180,44 +217,72 @@ impl TableFile {
...
-				len
-			},
-			Some((map, file)) => {
-				let new_len = try_io!(file.metadata()).len() + GROW_SIZE_BYTES;
-				try_io!(file.set_len(new_len));
-				if map.len() < new_len as usize {
-					let new_map = mmap(&file, new_len as usize)?;
-					let old_map = std::mem::replace(map, new_map);
-					try_io!(old_map.flush());
+		let map_and_file_len = map_and_file.len();
```
arkpar, line 220: `num_maps`

src/file.rs
```diff
@@ -180,44 +217,72 @@ impl TableFile {
 
 	pub fn grow(&self, entry_size: u16) -> Result<()> {
 		let mut map_and_file = self.map.write();
```
arkpar, line 219: `maps`

src/index.rs
cheme: great catch, thanks :+1:

src/ref_count.rs
```diff
@@ -77,8 +93,10 @@ fn total_chunks(index_bits: u8) -> u64 {
 	1u64 << index_bits
 }
 
-fn file_size(index_bits: u8) -> u64 {
-	total_entries(index_bits) * ENTRY_BYTES as u64 + META_SIZE as u64
+fn file_size(index_bits: u8, max_chunks: Option<u64>) -> u64 {
+	let max_size = max_chunks.map(|c| c * CHUNK_LEN as u64);
+	let total = total_entries(index_bits) * 8 + META_SIZE as u64;
```
MattHalpinParity, line 98: Should be:
`let total = total_entries(index_bits) * ENTRY_BYTES as u64 + META_SIZE as u64;`
Ref count entries are 16 bytes, not 8.

cheme: Indeed, awfull copy paste here.

cheme:
```suggestion
	let total = total_entries(index_bits) * ENTRY_BYTES as u64 + META_SIZE as u64;
```

src/file.rs
```diff
@@ -65,106 +74,134 @@ pub fn madvise_random(map: &mut memmap2::MmapMut) {
...
+					.disable_read_ahead()
+					.open(filepath.as_path()));
+				try_io!(disable_read_ahead(&file));
+				let len = try_io!(file.metadata()).len();
+				if len == 0 {
+					// Preallocate.
+					capacity += GROW_SIZE_BYTES / entry_size as u64;
+					try_io!(file.set_len(GROW_SIZE_BYTES));
+				} else {
+					capacity = len / entry_size as u64;
```
MattHalpinParity, line 138: Think this should be `+=`.  Though it doesn't cause any obvious problems as if the value is incorrect then grow is called and the correct capacity is calculated there.

cheme: Thanks, sure `+=` is right here.

PR #241, opened 2024-07-29
Propagate errors

MattHalpinParity:
As mentioned in Polkadot-sdk [#4939](https://github.com/paritytech/polkadot-sdk/issues/4939)
This change propagates all errors other than NotFound as that is needed to determine if the table file needs to be created or not.
@arkpar @cheme Does this seem reasonable?


cheme (approved):
works for me.
when looking at the issue I also thought that maybe `init_table_data` could return an error when `needs_free_entrie` is true (in this case the file should exists, if it doesn't there was some file loss or an initialization problem).

arkpar (approved):

MattHalpinParity:
> works for me. when looking at the issue I also thought that maybe `init_table_data` could return an error when `needs_free_entrie` is true (in this case the file should exists, if it doesn't there was some file loss or an initialization problem). (needs_free_entry is use with the order table where IIRC there is no lazy table init)

I think it should skip reading the file if it's missing so avoiding the None issue.

PR #242, opened 2024-10-03
Bump version to include multi tree changes

MattHalpinParity:
Bump version


cheme (approved):
I don't remember if there is changes to existing api, might just switch to 0.5 just in case.

arkpar (approved):

MattHalpinParity:
> I don't remember if there is changes to existing api, might just switch to 0.5 just in case.

Yup, I'll make it 0.5.0.

PR #250, opened 2026-05-08
feat(log): use ahash on live overlays

Klapeyron:
## Background

The three LogOverlay maps (index, value, ref_count) are keyed by u64 chunk-position indices, not user input — they originate from the internal value-table free list or from the high bits of blake2(column_key) (the column-level cryptographic key hash that sits above this layer).

The default std HashMap hasher (SipHash-1-3, per-process random seed) adds a second DoS-resistant layer as defence-in-depth. ahash::RandomState provides the same property — randomly seeded per process, so an attacker cannot precompute collisions — at ~3× lower CPU on short keys, using AES-NI rounds where available. Both are non-cryptographic in the formal sense; this is a "well-studied DoS hash" vs "newer DoS hash" trade.

On the warm-cache Db::get benchmark (8 reader threads, multi-column substrate-shape workload) this gives +15-22 % concurrent reader qps with no on-disk format change, no semantics change, and 36/36 lib tests passing.

The local LogWriter scratch maps keep their existing hashers — they're single-threaded and short-lived.

## Benchmarking

We made an experiment with live network sync with following results:

```
Validator	Start (UTC)	Finish (UTC, extrap.)	Duration	Avg rate
validator-04	16:41:51	18:03:38	1h 21m 47s	34.1 bps
validator-05	16:41:52	17:51:33	1h 09m 40s	40.0 bps
val-05 finished 12m 06s earlier than val-04.
```

sync_progress

## What this PR does

**File: `src/log.rs`**

- Import `std::collections::hash_map::RandomState` (as `StdBuildHasher`) and `std::hash::BuildHasher`.
- Add `LogWriterScratchBuildHasher = StdBuildHasher` for writer-local index and ref-count scratch maps, with a short doc note.
- Rename the `ahash` map builder alias to `OverlayHasher` (`ahash::RandomState`).
- Make `IndexLogOverlay`, `ValueLogOverlay`, and `RefCountLogOverlay` generic over `S: BuildHasher + Default`, defaulting to `OverlayHasher` (unchanged for `LogOverlays`).
- Use `IndexLogOverlay<LogWriterScratchBuildHasher>` and `RefCountLogOverlay<LogWriterScratchBuildHasher>` inside `LogChange` and `FlushedLog`.
- Replace the standalone `ValueLogOverlayLocal` struct with `pub type ValueLogOverlayLocal = ValueLogOverlay<BuildIdHash>`.
- Extend the module comments so they match the background above and describe which hasher is used where (live overlays vs scratch maps).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,6 +15,7 @@ bench = []
 arc = []
 
 [dependencies]
+ahash = "0.8"
 blake2 = "0.10.6"
 bytes = { version = "1.7.1", optional = true }
 crc32fast = "1.4.2"
--- a/src/log.rs
+++ b/src/log.rs
@@ -13,8 +13,9 @@ use crate::{
 };
 use std::{
 	cmp::min,
-	collections::{HashMap, VecDeque},
+	collections::{hash_map::RandomState as StdBuildHasher, HashMap, VecDeque},
 	convert::TryInto,
+	hash::BuildHasher,
 	io::{ErrorKind, Read, Seek, Write},
 	sync::atomic::{AtomicBool, AtomicU32, AtomicU64, Ordering},
 };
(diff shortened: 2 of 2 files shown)


arkpar:
Look good, thank you for the PR

arkpar:
Published as v0.5.5

